PR #61, opened 2021-04-12
Add docs publish CI job

mtreinish:
### Summary

This commit adds a new github workflow that will publish documentation
to https://qiskit.org/partners/ionq on new tag push events.


mtreinish:
Ok, I've removed the on hold label here because I've updated the trigger to only occur on tags. The first run uploaded it for us since the release is already live: https://qiskit.org/documentation/partners/ionq/

amilstead (approved):

PR #64, opened 2021-04-14
Pin qiskit version to avoid result processing errors

amilstead:
### Summary

Pin to `qiskit>=0.25.0` to avoid result model instantiation errors.

### Details and comments

Without this, you get something like: 

```
TypeError: __init__() got an unexpected keyword argument <some_keyword_arg>
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,2 @@
-qiskit-terra>=0.16
-qiskit>=0.25.0
+qiskit-terra>=0.17.0
 requests>=2.24.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 qiskit-terra>=0.16
+qiskit>=0.25.0
 requests>=2.24.0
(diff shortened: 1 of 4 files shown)


mjk:
Can we frame this as the IonQ provider requires version 0.25+? Vs not compatible with...

requirements.txt
```diff
@@ -1,2 +1,3 @@
 qiskit-terra>=0.16
+qiskit>=0.25.0
```
mtreinish, line 2: This isn't actually correct, you don't depend on the qiskit metapackage (like you don't need aer to use the ionq provider). If there is a compatibility issue here (although I'm not sure what the issue is) it would be with terra. You should just bump the line above to be `qiskit-terra>=0.17.0`

amilstead: My mistake! Looks like I had some confusion internally, this happens in results processing when unexpected keyword args are passed through. I'll update this to pin terra.

ColemanCollins (approved):
lgtm! Thanks @mtreinish on clarifications and help!

PR #67, opened 2021-04-20
fix and test against sumulator output with integers

ColemanCollins:
### Summary
Because the ionq simulator has some floating point weirdness, we re-normalize the results to make sure that all of the probabilities in the random sampling add up to one.

When there is one output state where p=1 in the simulator output, numpy freaks out when we're normalizing, because 1 is an integer and it's not cool with that.

forcibly casting all of the states to floats beforehand fixes this.


amilstead (approved):
lgtm

PR #68, opened 2021-04-21
Amilstead/fix runtime version reporting

amilstead:
### Summary

* Don't accidentally append directory local git information to qiskit-ionq's version.
* Try extra hard to not include test directories in a final package build.


ColemanCollins (approved):
you love to see it

PR #72, opened 2021-05-19
Bump terra version pin and docs

amilstead:
Bump version pin to `qiskit-terra>=0.17.4`, to prevent legacy aqua/framework bugs.


ColemanCollins (approved):
lgtm!

PR #75, opened 2022-04-01
ci/cd fixes

Cynocracy:
Started as part of https://github.com/Qiskit-Partners/qiskit-ionq/pull/74 to fix CI/CD


amilstead (approved):
We will need to make some notes in our next release that we have dropped support for 3.6.

Cynocracy:
Do you know where we would do that?

PR #80, opened 2022-04-18
Pipe target through the qiskit stack, add test

Cynocracy:
### Summary

This fixes a bug in the behavior for 0.2.0, which accepted more specific backend names, but did not pass them server-side

### Details and comments

Planning to release as 0.2.1 if there are no objections.

### DIFF
--- a/qiskit_ionq/ionq_backend.py
+++ b/qiskit_ionq/ionq_backend.py
@@ -26,6 +26,7 @@
 
 """IonQ provider backends."""
 
+import abc
 import warnings
 
 import dateutil.parser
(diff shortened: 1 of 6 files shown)


mtreinish:
One thing to keep in mind here is that in the v2 backend interface the name `target` is used as the attribute on the backend object containing a [`Target`](https://qiskit.org/documentation/stubs/qiskit.transpiler.Target.html#qiskit.transpiler.Target) object which is the QPU model used by the compiler, see: https://qiskit.org/documentation/apidoc/providers.html#migrating-between-backend-api-versions . So there might be some confusion when the ionq backend class is updated to use the newer interface if there is a `target` attribute and a `with_target` method which are completely unrelated.

Cynocracy:
Thanks! with_name probably makes more sense, the way I'm using it, will do a pass to avoid confusion

amilstead (approved):
lgtm

PR #83, opened 2022-04-20
Raise n_qubits to max IonQ provides across all QPU backends

Cynocracy:
If a particular circuit is run on hardware that does not support that specific size, it will be rejected server-side.

This is a bit of a cludge around dynamic backend fetching, once that is completed, we shouldn't be dependent on these SDK-specific caps.


Cynocracy:
Going to merge this as a pre-release version so that folks can test out the new cap.

PR #89, opened 2022-05-24
Retries, and handling invalid JSON in error body

Cynocracy:
### Summary
Adds retries for common 'server not available' error codes on methods typically used in offline settings (submit / retrieve jobs)

Also fixes a bug where invalid JSON in the response body would cause failures creating an exception.

### DIFF
--- a/qiskit_ionq/ionq_client.py
+++ b/qiskit_ionq/ionq_client.py
@@ -32,6 +32,23 @@ from . import exceptions
 from .helpers import qiskit_to_ionq
 
 
+# https://support.cloudflare.com/hc/en-us/articles/115003014512-4xx-Client-Error
+# "Cloudflare will generate and serve a 409 response for a Error 1001: DNS Resolution Error."
+# We may want to condition on the body as well, to allow for some GET requests to return 409 in
+# the future.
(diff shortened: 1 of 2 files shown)


qiskit_ionq/ionq_client.py
```diff
@@ -84,10 +103,14 @@ def submit_job(self, job) -> dict:
         as_json = qiskit_to_ionq(
             job.circuit, job.backend().name(), job.backend().gateset(), job._passed_args
         )
-        req_path = self.make_path("jobs")
-        res = requests.post(req_path, data=as_json, headers=self.api_headers)
-        if res.status_code != 200:
-            raise exceptions.IonQAPIError.from_response(res)
+        while True:
```
amilstead, line 106: Worth figuring out how to use [`retry`](https://pypi.org/project/retry/) here?

amilstead (commented):
Can you add some tests?

qiskit_ionq/ionq_client.py
Cynocracy: very yes, I hated writing manual retries, haha.

Will take a look at what some minimal testing looks like with retry as well.

amilstead (approved):
LGTM!

PR #94, opened 2022-06-04
Added User-Agent header [unitaryhack]

### LINKED ISSUE #91: Add a user-agent string
### What is the desired enhancement?
Every net-friendly tool should use user agents

Add a valid user-agent header that indicates the request is from qiskit_ionq, and a particular version

WingCode:
### Summary
Added User-Agent string.
closes #91 

### Details and comments
An observation, the naming of User-Agent strings is not [standardised](https://webaim.org/blog/user-agent-string-history/).

### DIFF
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -38,9 +38,9 @@ from qiskit import __version__ as qiskit_terra_version
 from qiskit.circuit import controlledgate as q_cgates
 from qiskit.circuit.library import standard_gates as q_gates
 
-from . import exceptions
 # Use this to get version instead of __version__ to avoid circular dependency.
 from importlib_metadata import version
+from . import exceptions
 
 # the qiskit gates that the IonQ backend can serialize to our IR
 # not the actual hardware basis gates for the system — we do our own transpilation pass.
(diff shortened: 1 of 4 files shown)


ColemanCollins:
Hey @WingCode, thanks for your PR. Per the original issue, one of the things we want in a successful PR here is a bit more information of the rest of the stack being used on submission, including versions. This allows for much more robust debugging and feature detection than just the provider's name.

I'm imagining a good solution here would include at least these three things:
- Provider Version (`qiskit-ionq/0.3.3`)
- Qiskit Terra Version (`qiskit-terra/0.20.2)
- Python Version (`python/3.8.9`)

ColemanCollins (changes requested):
Hey @WingCode, thanks for your PR. It's a great start. Per the original issue, one of the things we want in a successful PR here is a bit more information of the rest of the stack being used on submission, including versions. This allows for much more robust debugging and feature detection than just the provider's name.

I'm imagining a good solution here would include at least these three things:
- Provider Version (`qiskit-ionq/0.3.3`)
- Qiskit Terra Version (`qiskit-terra/0.20.2`)
- Python Version (`python/3.8.9`)

ColemanCollins:
n.b. it probably makes sense to do this as you're building the client instance, instead of on submission.

WingCode:
@ColemanCollins Thank you for the great inputs!

> I'm imagining a good solution here would include at least these three things:
> 
> * Provider Version (`qiskit-ionq/0.3.3`)
> * Qiskit Terra Version (`qiskit-terra/0.20.2`)
> * Python Version (`python/3.8.9`)

Can we also add the OS name to the user-agent string using `platform.system()` ?

ColemanCollins (changes requested):
Hey @WingCode, thanks for the great update. One minor cleanup thing and we're good here.

Also would appreciate if you added a quick test for this behavior (just confirm a generated request includes the substring of the package's own version or something) but this is simple enough that I'm not going to require it + want to make sure this gets in before Friday so you can get credit as part of unitaryHACK

ColemanCollins (approved):
Great stuff! thanks so much for your effort and contributions here

PR #96, opened 2022-06-07
Add job_settings to set arbitrary API fields for new features

Cynocracy:
Going to pre-release this to unblock some local folks

### DIFF
--- a/qiskit_ionq/ionq_backend.py
+++ b/qiskit_ionq/ionq_backend.py
@@ -125,7 +125,7 @@ class IonQBackend(Backend):
 
     @classmethod
     def _default_options(cls):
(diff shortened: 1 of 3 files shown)


qiskit_ionq/ionq_backend.py
```diff
@@ -125,7 +125,7 @@ class IonQBackend(Backend):
 
     @classmethod
     def _default_options(cls):
-        return Options(shots=1024)
+        return Options(shots=1024, api_settings=None)
```
gmauricio, line 128: What do you think if we rename to job_settings ? I think that'd be closer to the intent.

amilstead (approved):

PR #98, opened 2022-06-24
Allow setting custom headers

Cynocracy:


amilstead (approved):
If only we used dataclasses. 🤷

PR #102, opened 2022-09-14
Support setting noise model as an option to SimulatorBackend

Cynocracy:


qiskit_ionq/helpers.py
```diff
@@ -385,6 +384,11 @@ def qiskit_to_ionq(circuit, backend_name, gateset="qis", passed_args=None):
             "qiskit_header": qiskit_header,
         },
     }
+    if target == "simulator":
+        ionq_json["noise"] = {
+                "model": backend.options.noise_model,
+                "seed": backend.options.sampler_seed,
```
ColemanCollins, line 390: does sampler seed === noise seed? i.e. do we use that noise seed just for the shot sampling aspect of the noisy sim or also for some aspect of the noise generation?

it's probably fine to use one param here no matter what as the goal is the same; the internals of the simulator shouldn't change too much, and the idea of having a different noise seed and sampling seed is silly to begin with (esp because the API only wants one general-purpose seed). 

Just wondering if it makes semantic sense to change the name to a more generic`seed` or something (although that would be a breaking change...)

ColemanCollins (approved):
lgtm! comment about sampler_seed is not really blocking imo, just something to think about

qiskit_ionq/helpers.py
Cynocracy: ah, gotcha. I prefer using the same seed and just treating the whole process of providing results as 'sampling'

ColemanCollins: yeah; works for me. just thought I'd call it out for discussion

PR #103, opened 2022-10-12
Support new arbitrary angle param for MS Gate

gmauricio:
### Summary

MS Native Gate now supports an arbitrary theta angle.

### Details and comments

It will default to 0.25 (equivalent to half PI), which is equivalent to the previous implementation of this gate.

### DIFF
--- a/qiskit_ionq/ionq_gates.py
+++ b/qiskit_ionq/ionq_gates.py
@@ -106,7 +106,7 @@ class MSGate(Gate):
     .. math::
 
        MS(\phi_0, _\phi_1, \theta) q_0, q_1 =
-            \frac{1}{\sqrt{2}}\begin{pmatrix}
+            \begin{pmatrix}
                 cos{\th} & 0         & 0 & -i*e^{-i*2*\pi(\phi_0+\phi_1)}*sin{\th} \\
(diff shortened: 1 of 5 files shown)


qiskit_ionq/ionq_gates.py
```diff
@@ -105,40 +105,44 @@ class MSGate(Gate):
...
         )
 
     def __array__(self, dtype=None):
         """Return a numpy.array for the MS gate."""
         phi0 = self.params[0]
         phi1 = self.params[1]
-        diag = 1 / math.sqrt(2)
+        theta = self.params[2]
+        diag = numpy.cos(2 * math.pi * theta)
+        sin = numpy.sin(2 * math.pi * theta)
```
Cynocracy, line 138: could also call this adag for antidiagonal

test/helpers/test_qiskit_to_ionq.py
```diff
@@ -332,6 +332,8 @@ def test_full_native_circuit(simulator_backend):
     )
     registers = actual.pop("registers") or {}
 
+    print(actual)
```
Cynocracy, line 335: delete

Cynocracy (changes requested):

Cynocracy:
Looks good aside from the print! Once that's done will approve

Cynocracy (approved):

PR #106, opened 2022-11-30
Fix calibration() method

Cynocracy:
Also update to use the v0.2 base URL

There were a few issues with calibration(), mainly:

1. list calibrations endpoint no longer exists
2. code that was ported to raise_for_status was ported incorrectly, doubly breaking it (mea culpa on this one, I think)

This was tested manually, we'll have to add this to our internal integration testing to avoid this kind of breakage, which I can do as a followup


Cynocracy:
Ah, actually, it's probably better to use https://docs.ionq.com/#tag/characterizations/operation/getCharacterizationsCurrentForBackend

gmauricio (approved):

PR #107, opened 2023-02-08
upgrade to use API v0.3

gmauricio:
### Summary
Upgrade to v0.3 version of the IonQ API.

### Details and comments
- Results are no longer part of the job status response, but instead they have to be fetched from a new URL.
- Rsults won't come already mapped to the measured bits, but have to be mapped manually to the classical register.


qiskit_ionq/ionq_job.py
```diff
@@ -47,7 +47,31 @@
...
+
+    if not clbits:
+        return {}
+
+    mapped_output = {}
+
+    def get_bitvalue(bitstring, bit):
+        if bit is not None and 0 <= bit < len(bitstring):
+            return bitstring[bit]
+        return '0'
```
Cynocracy, line 61: Should we fail instead of returning 0 outside the bounds?

qiskit_ionq/ionq_job.py
```diff
@@ -47,7 +47,31 @@
...
+
+    for value, probability in data.items():
+        bitstring = bin(int(value))[2:].rjust(num_qubits, "0")[::-1]
```
Cynocracy, line 64: meas_bitstring maybe?

qiskit_ionq/ionq_job.py
```diff
@@ -47,7 +47,31 @@
...
+
+        bitvalue = int(''.join([get_bitvalue(bitstring, bit) for bit in clbits])[::-1], 2)
```
Cynocracy, line 66: outvalue ?

Cynocracy (approved):

qiskit_ionq/ionq_job.py
gmauricio: Since at the end this comes from Qiskit measurement instructions I wonder if they validate this already 🤔 
Still failing could be a good idea, I'll take a look

Cynocracy: Ah gotcha. Not to worried about it in that case :+1:

qiskit_ionq/ionq_client.py
```diff
@@ -39,7 +39,7 @@ class IonQClient:
     """IonQ API Client
 
     Attributes:
-        _url(str): A URL base to use for API calls, e.g. ``"https://api.ionq.co/v0.1"``
```
Cynocracy, line 42: We probably want to do a minor or major semver bump in qiskit

I think we'll need to say retrieve_job no longer populates histogram

Cynocracy: Oh, actually, I don't think we expose that class (nice)

Cynocracy: https://qiskit.org/documentation/partners/ionq/apidocs/backends.html ah, It's in backend.

Cynocracy: All good, we expose sane wrappers in all cases I can see on https://qiskit.org/documentation/partners/ionq/apidocs/backends.html

PR #109, opened 2023-02-13
add error_mitigation settings

gmauricio:
### Summary
Add error_mitigation settings

### Details and comments
Includes:
- Settings on job submission to enable/disable and configure symmetrization
- New results parameter to fetch results by different aggregation methods if job was symmetrized

### DIFF
--- a/qiskit_ionq/constants.py
+++ b/qiskit_ionq/constants.py
@@ -63,4 +63,19 @@ class JobStatusMap(enum.Enum):
     FAILED = jobstatus.JobStatus.ERROR.name
 
 
+class AggregationType(enum.Enum):
+    """Class for job results aggregation enumerated type."""
+
+    AVERAGE = 'average'
+    MAJORITY = 'majority'
+
+
+class ErrorMitigation(enum.Enum):
+    """Class for error mitigation settings enumerated type."""
+
+    NO_SYMMETRIZATION = {"symmetrization": False}
+    SYMMETRIZATION_AVERAGE = {"symmetrization": {"aggregation": "average"}}
+    SYMMETRIZATION_MAJORITY = {"symmetrization": {"aggregation": "majority"}}
+
+
 __all__ = ["APIJobStatus", "JobStatusMap"]
--- a/qiskit_ionq/constants.py
+++ b/qiskit_ionq/constants.py
@@ -67,15 +67,14 @@ class AggregationType(enum.Enum):
     """Class for job results aggregation enumerated type."""
 
     AVERAGE = 'average'
-    MAJORITY = 'majority'
+    PLURALITY = 'plurality'
 
 
 class ErrorMitigation(enum.Enum):
     """Class for error mitigation settings enumerated type."""
 
+    SYMMETRIZATION = {"symmetrization": True}
     NO_SYMMETRIZATION = {"symmetrization": False}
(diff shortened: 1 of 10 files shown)


qiskit_ionq/constants.py
```diff
@@ -63,4 +63,18 @@ class JobStatusMap(enum.Enum):
     FAILED = jobstatus.JobStatus.ERROR.name
 
 
+class AggregationType(enum.Enum):
```
Cynocracy, line 66: I'm meh on having this defined in the client.

qiskit_ionq/ionq_job.py
```diff
@@ -255,9 +256,10 @@ def result(self):
         Returns:
             Result: A Qiskit :class:`Result <qiskit.result.Result>` representation of this job.
         """
-        # Short-circuit if we have already cached the result for this job.
-        if self._result is not None:
-            return self._result
+        # TODO: cache results by aggregation type
+
+        if aggregation is not None and not isinstance(aggregation, AggregationType):
```
Cynocracy, line 261: Can we warn instead of throw here?

Cynocracy (changes requested):

qiskit_ionq/ionq_job.py
gmauricio: why not throw ? the request would fail otherwise.

qiskit_ionq/constants.py
```diff
@@ -63,4 +63,18 @@ class JobStatusMap(enum.Enum):
...
+
+    AVERAGE = 'average'
+    PLURALITY = 'plurality'
+
+
+class ErrorMitigation(enum.Enum):
+    """Class for error mitigation settings enumerated type."""
+
+    SYMMETRIZATION = {"symmetrization": True}
+    NO_SYMMETRIZATION = {"symmetrization": False}
```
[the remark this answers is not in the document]
gmauricio: it disables it, independently from whatever is default

qiskit_ionq/constants.py
gmauricio: what do you mean?

Cynocracy: We might add more aggregation settings, it's weird to check that they're in an allowed list that's part of the client, because that would block new settings behind a client upgrade.

qiskit_ionq/constants.py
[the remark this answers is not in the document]
Cynocracy: Can we add a DEFAULT that lets you leave it unset?

qiskit_ionq/ionq_job.py
Cynocracy: Relying on server-side validation means we won't need to update the client when we add a new aggregation type.

Same issue with doing target totally client side, you either need to wire up to an endpoint that lets you know the available options (for targets, /backends), or you're heavily restricting what the client can do for something that should be checked server-side anyway.

gmauricio: Got it. Makes sense.

qiskit_ionq/constants.py
splch: I think sending an aggregation then returning a failed run with the error message specifying which aggregations are supported would be nice. we could then define those aggregations in openapi docs?

PR #111, opened 2023-03-12
Include full http response on exception

splch:
### Summary

Includes full http response in expections

### Details and comments

includes this `http_response` [alongside](https://github.com/Qiskit-Partners/qiskit-ionq/blob/c757237c76bacf31ebf3391f7dd1ee86c57c5b23/qiskit_ionq/exceptions.py#L157-L159) `error_type`. alternatively we can just show the `http_response`.

### DIFF
--- a/qiskit_ionq/exceptions.py
+++ b/qiskit_ionq/exceptions.py
@@ -122,7 +122,8 @@ class IonQAPIError(IonQError):
         """
         # TODO: Pending API changes will cleanup this error logic:
         status_code = response.status_code
-        http_response = response.reason
+        headers = response.headers
+        body = response.text
         try:
             response_json = response.json()
         except jd.JSONDecodeError:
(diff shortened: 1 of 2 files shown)


test/test_exceptions.py
```diff
@@ -47,24 +48,29 @@ def test_gate_error_str_and_repr():
...
     assert err.status_code == 500
     assert err.error_type == "internal_error"
 
 
 def test_api_error_str_and_repr():
     """Test that IonQAPIError has a str/repr that includes args."""
-    err = exceptions.IonQAPIError("an error", 500, "internal_error")
-    expected = "IonQAPIError(message='an error',status_code=500,error_type='internal_error')"
+    err = exceptions.IonQAPIError(
+        "an error", "http response", 500, "internal_error")
```
Cynocracy, line 69: can you try this with a real requests.Response to see if it serializes OK?

qiskit_ionq/exceptions.py
```diff
@@ -118,10 +122,11 @@ def from_response(cls, response):
         """
         # TODO: Pending API changes will cleanup this error logic:
         status_code = response.status_code
+        http_response = response.reason
```
Cynocracy, line 125: Reason just returns the text form of the http code, yeah?
https://requests.readthedocs.io/en/latest/api/#requests.Response.reason

I think it's an improvement, but could we also log response headers and the full body? (for example we strip parts of the body if code is found below, and that might be useful for debugging as well)

splch: sounds good i included the headers and full body

test/test_exceptions.py
splch: alright im now using requests.request and response so it's pretty realistic

test/test_exceptions.py
```diff
@@ -47,118 +49,176 @@ def test_gate_error_str_and_repr():
...
     """Test that IonQAPIError can be made directly from a response JSON dict."""
-    fake_response = mock.MagicMock()
-    fake_response.json = mock.MagicMock(
+    # Create a response object
+    response = requests.Response()
+    response.status_code = 500
+    response.headers = {"Content-Type": "text/html"}
+    response._content = b"<!doctype html><html><body>Hello IonQ!</body></html>"
+    # Mock the json method of the response object
+    response.json = mock.MagicMock(
```
Cynocracy, line 100: you might be able to avoid mocking this if you put a json string into _content (though that seems a bit odd too. It's test only code so I don't mind much, but we could use https://github.com/getsentry/responses if you're interested in trying it out)

Cynocracy (approved):

Cynocracy:
Looks great!

PR #113, opened 2023-03-12
use ionq_native_basis_gates in transpile

splch:
### Summary

Uses Qiskit's `basis_gates` argument with `transpile` to have gate parity between circuit and computer.


splch:
replaced by #166

PR #114, opened 2023-03-16
Warn when a Qiskit circuit has null mappings for all registers

splch:
### Summary

Emits a warning when a submitted circuit has no valid qubit->bit mappings.

### Details and comments

```python
provider = IonQProvider()

qc = qiskit.QuantumCircuit(1, 1, name="null circuit")
qc.h(0)

job = backend.run(qc)

> /qiskit-ionq/qiskit_ionq/ionq_backend.py:359: UserWarning: Circuit null mappings is not measuring any qubits
```

### DIFF
--- a/qiskit_ionq/ionq_backend.py
+++ b/qiskit_ionq/ionq_backend.py
@@ -180,7 +180,8 @@ class IonQBackend(Backend):
             ) from ex
 
         if url is None:
-            raise exceptions.IonQCredentialsError("Credentials `url` may not be None!")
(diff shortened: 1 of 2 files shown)


Cynocracy:
I would actually slightly prefer that this warning be done server-side through https://github.com/Qiskit-Partners/qiskit-ionq/blob/39989e51f3970a9ee11416ffa4f26a051bcc62bf/qiskit_ionq/ionq_job.py#L346-L348 but I am okay with one being added client side as well

PR #116, opened 2023-03-20
Adds noise_model to override in run method

splch:
### Summary

Follow [Qiskit convention](https://qiskit.org/documentation/stubs/qiskit_aer.AerSimulator.run.html#qiskit-aer-aersimulator-run) of specifying options in `run(..., **kwargs)`.

### Details and comments

Can invoke with

```python
provider = IonQProvider()

job = backend.run(qc, noise_model="ideal")

print(job.get_counts())
```

### DIFF
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -395,7 +395,7 @@ def qiskit_to_ionq(circuit, backend, passed_args=None):
     }
     if target == "simulator":
         ionq_json["noise"] = {
-            "model": passed_args.get("noise_model", backend.options.noise_model),
+            "model": passed_args.get("noise_model") or backend.options.noise_model,
             "seed": backend.options.sampler_seed,
         }
     settings = passed_args.get("job_settings") or None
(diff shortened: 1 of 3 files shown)


qiskit_ionq/helpers.py
```diff
@@ -389,8 +395,8 @@ def qiskit_to_ionq(circuit, backend, passed_args=None):
     }
     if target == "simulator":
         ionq_json["noise"] = {
-                "model": backend.options.noise_model,
-                "seed": backend.options.sampler_seed,
+            "model": passed_args.get("noise_model", backend.options.noise_model),
```
amilstead, line 398: Seems to be the most significant material change here. Only word of caution/nit-suggestion is to: `passed_args.get("noise_model", None) or backend.options.noise_model)`, to protect against cases where `passed_args["noise_model"]` exists but is for Falsey or otherwise invalid.

amilstead (approved):
lgtm. @Cynocracy anything else to add here?

qiskit_ionq/helpers.py
splch: Sounds good I'll update that 👍

amilstead (approved):

PR #124, opened 2023-05-05
rename debiasing + backend

splch:
### Summary

Renames symmetrization to debiasing, updates the v0.3 backend naming scheme, and adds a `extra_request_payload` for specifying post / get parameters

### Details and comments

Screenshot 2023-05-09 at 9 18 11 AM

### DIFF
--- a/qiskit_ionq/constants.py
+++ b/qiskit_ionq/constants.py
@@ -73,8 +73,8 @@ class AggregationType(enum.Enum):
 class ErrorMitigation(enum.Enum):
     """Class for error mitigation settings enumerated type."""
 
-    SYMMETRIZATION = {"symmetrization": True}
-    NO_SYMMETRIZATION = {"symmetrization": False}
+    DEBIASING = {"debias": True}
+    NO_DEBIASING = {"debias": False}
 
 
 __all__ = ["APIJobStatus", "JobStatusMap"]
--- a/qiskit_ionq/constants.py
+++ b/qiskit_ionq/constants.py
@@ -63,13 +63,6 @@ class JobStatusMap(enum.Enum):
     FAILED = jobstatus.JobStatus.ERROR.name
 
 
-class AggregationType(enum.Enum):
-    """Class for job results aggregation enumerated type."""
-
-    AVERAGE = "average"
-    PLURALITY = "plurality"
-
-
 class ErrorMitigation(enum.Enum):
     """Class for error mitigation settings enumerated type."""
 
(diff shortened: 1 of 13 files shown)


qiskit_ionq/constants.py
```diff
@@ -63,18 +63,11 @@ class JobStatusMap(enum.Enum):
...
-    PLURALITY = "plurality"
-
-
 class ErrorMitigation(enum.Enum):
     """Class for error mitigation settings enumerated type."""
 
-    SYMMETRIZATION = {"symmetrization": True}
-    NO_SYMMETRIZATION = {"symmetrization": False}
+    DEBIASING = {"debias": True}
+    NO_DEBIASING = {"debias": False}
```
Cynocracy, line 70: I don't like that this requires SDK-side updates if we offer new error mitigation settings.

Is it possible to pass in an arbitrary JSON object for error_mitigation ? Can we support that?

splch: im fine with that but then we need to have a source of clear documentation of error mitigation settings that we support - that could be https://docs.ionq.com/#tag/jobs under `error_mitigation`. ill add that as well to the readme

mocha: Ping me on the docs change if you don't mind; mostly curious on the process for rolling those out

qiskit_ionq/helpers.py
```diff
@@ -344,7 +343,7 @@ def decompress_metadata_string_to_dict(input_string):  # pylint: disable=invalid
     return json.loads(decompressed)
 
 
-def qiskit_to_ionq(circuit, backend, passed_args=None):
+def qiskit_to_ionq(circuit, backend, passed_args=None, _params=None):
```
Cynocracy, line 346: _params is a bit awkwardly named, imo, it makes me think QueryParams, and the underscore seems to indicate you shouldn't actually use it (_ stuff in python is usually 'private')

Can we name it like, extra_request_payload or something like that?

Sorry to nitpick so much on this bit

qiskit_ionq/ionq_client.py
```diff
@@ -186,25 +196,30 @@ def get_calibration_data(self, backend_name: str) -> dict:
...
 
         return res.json()
 
     @retry(exceptions=IonQRetriableError, max_delay=60, backoff=2, jitter=1)
-    def get_results(self, job_id: str, aggregation=None):
+    def get_results(
+        self,
+        job_id: str,
+        sharpen: Optional[bool] = None,
+        extra_request_payload: Optional[dict] = None,
```
Cynocracy, line 212: ah, sorry... when I saw that this was updating a dict I thought it was part of the jobs payload. extra_query_params does seem more appropriate (sorry to back and forth so much, this field being optional means folks will need to use its name to reference it so I'm a bit fixated on it at the moment).

Cynocracy (approved):

PR #125, opened 2023-05-10
Remove outdated docs template overrides

Eric-Arellano:
### Summary

These template overrides mean that you won't be getting improvements to qiskit_sphinx_theme. They also risk breaks because we assume the files are not overridden. It can be valid to override the templates, but should only be done intentionally and with an understanding that increases your maintenance burden.

This also:

* Adds missing `pyproject.toml`, which is strongly encouraged to have for Python packaging.
* Activates jQuery, which is no longer activated by default.
* Removes sphinx-rtd-theme install, which isn't used.
* Removes `html_theme_path` call, which was not necessary and is deprecated.
*  Pins the qiskit_sphinx_theme by minor version. While we try to keep breaking changes to a minimum, we are undergoing big improvements to the theme right now so this gives you better stability.

Warning: I could not get the docs to build locally.


Eric-Arellano:
Thanks for your help earlier @Cynocracy! Bump on this one. 

Once we merge it, could you please trigger a docs deploy? It looks like the theme is out of date. After this PR is merged, you can do that by going to the Actions tab, navigating to the correct workflow name, then clicking the run button:

Screenshot 2023-05-23 at 12 17 41 PM

Cynocracy:
No problem! Sure, I can shepherd this to a docs release

PR #131, opened 2023-06-22
default ionq_qpu to ionq_qpu.harmony

splch:
### Summary

make documentation using `ionq_qpu` as backend work

### Details and comments

defaults ionq_qpu to harmony.

todo: use cheapest available option as default


dwiddows (approved):
Approved with conversation suggesting that we check that the prefix is `ionq_` before removing it.

PR #132, opened 2023-06-22
Update version.py to 0.4.1

splch:
### Summary

fixes common issue from documentation

### Details and comments

asap release

### DIFF
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -33,6 +33,7 @@ import json
 import gzip
 import base64
 import platform
+import warnings
 
 from qiskit import __version__ as qiskit_terra_version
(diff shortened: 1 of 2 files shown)


Cynocracy:
I'm afraid if we add a silently upgrade to harmony it will be hard to notice that it's happening / it might cause further issues down the line if we decomission harmony, or if we do end up using an 'ionq_qpu' meta-target again.

Can we print a warning?

splch:
> I'm afraid if we add a silently upgrade to harmony it will be hard to notice that it's happening / it might cause further issues down the line if we decomission harmony, or if we do end up using an 'ionq_qpu' meta-target again.
> 
> Can we print a warning?

got it thanks!

dwiddows (approved):

PR #135, opened 2023-07-17
Allow for additional metadata to be specified

splch:
### Summary

Add an extra_metadata argument to specify / filter on additional information

### Details and comments

adds extra metadata + upgrades python testing to 3.10

### DIFF
--- a/qiskit_ionq/ionq_client.py
+++ b/qiskit_ionq/ionq_client.py
@@ -97,6 +97,7 @@ class IonQClient:
             job.backend(),
             job._passed_args,
             job.extra_query_params,
+            job.extra_metadata,
         )
         req_path = self.make_path("jobs")
         res = requests.post(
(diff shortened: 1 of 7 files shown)


qiskit_ionq/ionq_client.py
```diff
@@ -97,6 +97,7 @@ def submit_job(self, job) -> dict:
             job.backend(),
             job._passed_args,
             job.extra_query_params,
+            job.extra_metadata,
```
Cynocracy, line 100: thoughts on a mode where this comes by default from the passsed in QuantumCircuit?

Cynocracy (approved):

qiskit_ionq/ionq_client.py
splch: got it thanks for the suggestion!

PR #138, opened 2023-08-03
is ideal means no sampler

splch:
### Summary

fixes boolean bug

### Details and comments

makes ideal simulator run by default

### DIFF
--- a/qiskit_ionq/ionq_job.py
+++ b/qiskit_ionq/ionq_job.py
@@ -427,7 +427,7 @@ class IonQJob(JobV1):
                 self._num_qubits,
                 self._clbits,
                 shots,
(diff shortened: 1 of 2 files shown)


Cynocracy (changes requested):
The sampling from the ideal simulator was to be able to provide counts, it's a bit strange but this is the intention of the logic

The noisy simulator already measures server side and so should not be sampled from

splch:
The IonQ simulator API natively produces ideal probabilities based on a statevector simulation, not counts. For ease of interoperability and portability, this package uses a simple random sampling to produce "counts," and also exposes the true probabilities. The code samples below show this in more detail.

PR #142, opened 2023-09-29
Fixup serialization of ZZ gate

Cynocracy:
Logic was assuming that for gates with 1 input that input was 'phase' which is not true for ZZ, which just has 'theta'


splch (approved):

Cynocracy:
ty :pray:

PR #145, opened 2023-10-24
raise IonQRetriableError when get fails

splch:
### Summary

try-catch block has been added around the network request to handle exceptions that can be retried

### DIFF
--- a/qiskit_ionq/ionq_client.py
+++ b/qiskit_ionq/ionq_client.py
@@ -78,6 +78,31 @@ class IonQClient:
         """
         return "/".join([self._url] + list(parts))
 
+    def _get_with_retry(self, req_path, params=None, headers=None, timeout=30):
+        """Make a GET request with retry logic and exception handling.
+
+        Args:
+            req_path (str): The URL path to make the request to.
+            params (dict, optional): Parameters to include in the request.
+            headers (dict, optional): Headers to include in the request.
+            timeout (int, optional): Timeout for the request.
+
+        Raises:
+            IonQRetriableError: When a retriable error occurs during the request.
+
+        Returns:
+            Response: A requests.Response object.
+        """
+        try:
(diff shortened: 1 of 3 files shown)


qiskit_ionq/ionq_client.py
```diff
@@ -244,7 +246,12 @@ def get_results(
             params.update(extra_query_params)
 
         req_path = self.make_path("jobs", job_id, "results")
-        res = requests.get(req_path, params, headers=self.api_headers, timeout=30)
```
Cynocracy, line 247: Can we also add to other requests.get calls in this file?

Ty!

Cynocracy (approved):

qiskit_ionq/ionq_client.py
splch: didnt mean to delete that haha nice catch

qiskit_ionq/ionq_client.py
```diff
@@ -77,6 +77,32 @@ def make_path(self, *parts):
...
+            IonQRetriableError: When a retriable error occurs during the request.
+
+        Returns:
+            Response: A requests.Response object.
+        """
+        try:
+            res = requests.get(
+                req_path, params=params, headers=headers, timeout=timeout
+            )
+            res.raise_for_status()
```
Cynocracy, line 100: You shouldn't call this here as the ionq specific raise_for_status is called later

qiskit_ionq/ionq_client.py
```diff
@@ -77,6 +77,32 @@ def make_path(self, *parts):
...
+        except requests.exceptions.RequestException as req_exc:
```
Cynocracy, line 101: if you remove raise_for_status you can just catch any error here, as they should all relate to the socket i/o

PR #154, opened 2023-11-13
support multi-circuit jobs

splch:
### Summary

enable multi-circuit submission and retrieval of results

```python
from qiskit import QuantumCircuit, transpile
from qiskit_ionq import IonQProvider
from math import pi

provider = IonQProvider()

qc1 = QuantumCircuit(2, 2, name="2 registers")
qc1.u(pi, pi / 2, pi / 4, 0)
qc1.measure_all()
qc1 = transpile(qc1, backend)

qc2 = QuantumCircuit(3, 3, name="inverted measurements")
qc2.h(1)
qc2.cx(1, 2)
qc2.measure([1, 2], [2, 0])

single_job = backend.run(qc1)
multi_job = backend.run([qc1, qc2])

single_job.get_counts()
# {'00 01': 1024}

multi_job.get_counts()
# [{'00 01': 1024}, {'000': 520, '110': 504}]

multi_job.get_counts(qc1)
# {'00 01': 1024}
```

### Notes

we should offer a `get_children()` method that returns child ids and associate with indices of the [Results' results](https://docs.quantum.ibm.com/api/qiskit/qiskit.result.Result#results) list.

look into how [experiments](https://qiskit-extensions.github.io/qiskit-experiments/tutorials/intro.html#what-is-qiskit-experiments) work

### Details and comments

- [x] submit multi-circuit
- [x] retrieve results
- [x] pass tests
- [x] add multi-circuit tests
- [x] handle list of circuit metadata
- [x] correctly order results and metadata


Cynocracy (approved):
Generally I think we need to handle the registers being keyed by job, I'd be okay with merging this and doing that in a separate PR, as we've discussed offline, but having the correct registers per job in a multicircuit submission will be important for reading the results correctly, imo

qiskit_ionq/helpers.py
```diff
@@ -363,15 +363,25 @@ def qiskit_to_ionq(
...
-            "memory_slots": circuit.num_clbits,  # int
-            "global_phase": circuit.global_phase,  # float
-            "n_qubits": circuit.num_qubits,  # int
-            "name": circuit.name,  # str
+            "memory_slots": circ.num_clbits,  # int
+            "global_phase": circ.global_phase,  # float
+            "n_qubits": circ.num_qubits,  # int
+            "name": circ.name,  # str
             # list of [str, int] tuples cardinality memory_slots
             "creg_sizes": creg_sizes,
```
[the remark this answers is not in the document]
splch: we could gzip them in the future

qiskit_ionq/ionq_job.py
```diff
@@ -413,36 +415,46 @@ def _format_result(self, data):
...
+            # to handle ionq returning different data structures for single and multiple circuits
+            if self._num_circuits > 1:
+                data = list(data.values())
+            else:
+                data = [data]
+            for idx in range(self._num_circuits):
+                (counts, probabilities) = _build_counts(
+                    data[idx],
+                    self._num_qubits,
+                    self._clbits,
```
Cynocracy, line 438: I think this will need to come from metadata as well,

Currently we set this into a top level "registers" field on the Job but this may vary from circuit to circuit.

qiskit_ionq/ionq_job.py
```diff
@@ -180,18 +176,22 @@ def __init__(
...
+        # Handle both single and list of circuits
         if circuit is not None:
-            self.circuit = circuit
+            if isinstance(circuit, (list, tuple)):
+                self.circuit = circuit
+                for circ in circuit:
+                    self.extra_metadata.update(circ.metadata or {})
+            else:
+                self.circuit = circuit
+                self.extra_metadata.update(circuit.metadata or {})
```
Cynocracy, line 194: This means that metadata from a circuit later in the list will override metadata earlier , yeah?

Do we want to warn/fail if that happens?

Cynocracy (changes requested):

qiskit_ionq/ionq_job.py
kielpins-qctrl: Hope it's ok for me to comment. From my POV as a user, I rely on accurate preservation of the circuit metadata all the way through to the results, as a feature of `qiskit_ibm_runtime`. Here's my use case.

Since queue times can be hours or days, I run a Python process to launch a multicircuit job, then I terminate it so it doesn't block during queueing time. Periodically, I check in manually on the job completion status. Then when the job completes, I run a separate Python process to analyse the results.

Now, how to keep track of the functionality of the various circuits in the job? Locally storing metadata and correlating it by job ID and circuit index strikes me as a fragile solution. Instead, I rely on the circuit metadata being accurately passed through to the results. Then each circuit result has the original metadata already attached, and I just parse it as part of the result object.

splch: thanks

> Hope it's ok for me to comment. From my POV as a user, I rely on accurate preservation of the circuit metadata all the way through to the results, as a feature of `qiskit_ibm_runtime`. Here's my use case.
> 
> Since queue times can be hours or days, I run a Python process to launch a multicircuit job, then I terminate it so it doesn't block during queueing time. Periodically, I check in manually on the job completion status. Then when the job completes, I run a separate Python process to analyse the results.
> 
> Now, how to keep track of the functionality of the various circuits in the job? Locally storing metadata and correlating it by job ID and circuit index strikes me as a fragile solution. Instead, I rely on the circuit metadata being accurately passed through to the results. Then each circuit result has the original metadata already attached, and I just parse it as part of the result object.

thats great context for using metadata :) thank you! ill try to fix this in the pr, otherwise itll come in a patch release soon.

kielpins-qctrl: Thanks for your kind response @splch. Can I ask about the timeline for this PR? Multicircuit job submission is currently a blocker for Q-CTRL's collaboration with John Gamble's team.

qiskit_ionq/ionq_job.py
```diff
@@ -410,39 +410,60 @@ def _format_result(self, data):
...
-                "counts": counts,
-                "probabilities": probabilities,
-                # Qiskit/experiments relies on this being present in this location in the
-                # ExperimentData class.
-                "metadata": qiskit_header or {},
+        job_result = [
+            {
+                "data": {},
+                "shots": shots,
+                "header": qiskit_header[i] or {},
```
splch, line 425: I'm not certain the ordering of results and metadata is consistent

qiskit_ionq/helpers.py
```diff
@@ -296,27 +296,30 @@ def get_register_sizes_and_labels(registers):
...
         str: encoded string
 
     """
-    serialized = json.dumps(metadata_dict)
+    serialized = json.dumps(metadata)
     compressed = gzip.compress(serialized.encode("utf-8"))
     encoded = base64.b64encode(compressed)
     encoded_string = encoded.decode()
+    encoded_string_length = len(encoded_string)
+    if encoded_string_length >= 40000:  # 40000 char is an IonQ API limitation
```
Cynocracy, line 317: slight preference for us doing this enforcement server-side only (notifying the user if the job submission is rejected) but I'm okay with this for now :)

qiskit_ionq/ionq_job.py
```diff
@@ -180,18 +176,22 @@ def __init__(
...
             self.extra_metadata = {}
             self._passed_args = {"shots": 1024, "sampler_seed": None}
 
+        # Handle both single and list of circuits
         if circuit is not None:
-            self.circuit = circuit
+            if isinstance(circuit, (list, tuple)):
+                self.circuit = circuit
+                for circ in circuit:
+                    self.extra_metadata.update(circ.metadata or {})
```
Cynocracy, line 191: Do we still want to combine all metadata at the top level like this?

qiskit_ionq/ionq_job.py
Cynocracy: Maybe serialize as circuit-hash like you first thought?

qiskit_ionq/helpers.py
splch: youre right we already raise an exception! ill remove this

qiskit_ionq/ionq_job.py
splch: ah we probably shouldnt to avoid going over the character limit! ill remove this too

Cynocracy (approved):

qiskit_ionq/ionq_job.py
splch: ill try using the circuit name first hopefully that works

karamsg:
Hi!
I think this PR introduced a bug in showing/parsing the result.

try the following on version 0.5.0 and any version afterwards: 
```
qc = QuantumCircuit(3)
qc.x(0)
qc.x(1)
qc.z(2)

cr = ClassicalRegister(2)
qc.add_register(cr)

qc.measure([1, 2], [0, 1])

provider = qiskit_ionq.IonQProvider(token="...")
backend = provider.get_backend("simulator")

res = backend.run(qc).result().get_counts()
# {'01': 1024} <- this is the correct answer (v0.5.0)
# {'11': 1024} <- this is what we get now (>0.5.0)
```

Cynocracy:
Hi @karamsg thanks for the report! We will take a look, https://github.com/qiskit-community/qiskit-ionq/pull/178 is another PR that touched similar functionality recently.

PR #155, opened 2023-12-18
Add SafeEncoder class to handle non-JSON-safe objects

splch:
### Summary

use a safeencoder to parse any object passed in the job

### Details and comments

uses the __string__ method if present otherwise calls repr

### DIFF
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -454,12 +454,19 @@ class SafeEncoder(json.JSONEncoder):
     """
 
     def default(self, o):
-        try:
-            return super().default(o)
-        except TypeError:
-            return (
-                str(o) if "__str__" in vars(type(o)) else repr(o)
-            )  # convert non-JSON-safe objects to string
+        funcs = [
+            lambda: super().default(o),
+            lambda: str(o),
+            lambda: repr(o),
+        ]
+
+        for func in funcs:
+            try:
+                return func()
+            except Exception as e:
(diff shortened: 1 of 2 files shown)


qiskit_ionq/helpers.py
```diff
@@ -448,6 +448,20 @@ def get_user_agent():
...
+class SafeEncoder(json.JSONEncoder):
+    """
+    Custom JSON encoder that handles non-JSON-safe objects by converting them to strings.
+    """
+
+    def default(self, o):
+        try:
+            return super().default(o)
+        except TypeError:
+            return (
```
Cynocracy, line 460: Can you log the fieldname when this happens?

Cynocracy (changes requested):

qiskit_ionq/helpers.py
```diff
@@ -448,6 +448,20 @@ def get_user_agent():
...
+                str(o) if "__str__" in vars(type(o)) else repr(o)
```
Cynocracy, line 461: would prefer another try catch here, maybe you can set it up as a for loop with a few options for the source of the string?

something like

funcs = [ lambda o: super().default(o), ...]

for func in funcs:
   try:
      return func()
   except:
       log
return 'unknown'

splch: great yeah i updated to do that - ill look into adding the field name now

Cynocracy:
LGTM with the changes

PR #159, opened 2024-01-18
Remove client-side metadata length check

Cynocracy:
We're raising this limit considerable server-side so we expect both not to hit the limit, and to have a reasonable error message from the API if it does so. 

Either way, the client checking the limit is no longer necessary and impedes further raising the limit in a way that benefits all existing users.


amilstead:
I am personally super stoked to see this.

splch (approved):

PR #165, opened 2024-02-08
Add `VirtualZGate` to native gateset

splch:
### Summary

adds a `VirtualZGate` to the native gateset as `gz`

### Details and comments

definition here: https://ionq.com/docs/getting-started-with-native-gates#single-qubit-gates


Cynocracy (changes requested):
Would like to see support for removing GZ gates from the final circuit before submission (via merging / omission at the end) or for us to support this in the API itself before merge

PR #171, opened 2024-02-29
add backend cancel job methods and save metadata

splch:
### Summary

```python
# Print all metadata
print(job._metadata)
```

```python
# Cancel jobs with the backend
backend.cancel_job('9a36b191-d8be-4b77-abcc-ec1ef0ac15d2')
backend.cancel_jobs(['6fb62d08-4e55-4fbd-8874-b01453290730', '37635b2b-994b-42f4-8d65-c4133328f95d'])
```

plus remove qiskit dep (warn if not installed)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-qiskit>=0.4.0
+qiskit>=0.46.0
 decorator>=5.1.0
 requests>=2.24.0
 retry>=0.9.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
-qiskit>=0.46.0
 decorator>=5.1.0
 requests>=2.24.0
 retry>=0.9.0
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -42,13 +42,14 @@
 
 """Sphinx doc build configuration."""
 import qiskit_sphinx_theme
+from qiskit_ionq.version import VERSION_INFO
 
 # -- Project information -----------------------------------------------------
 
 project = "Qiskit IonQ Provider"  # pylint: disable=invalid-name
 copyright = "2020, IonQ, Inc."  # pylint: disable=invalid-name,redefined-builtin
 author = "IonQ, Inc."  # pylint: disable=invalid-name
-release = "0.0.1"  # pylint: disable=invalid-name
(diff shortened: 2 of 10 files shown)


Cynocracy (approved):

requirements.txt
```diff
@@ -1,4 +1,4 @@
-qiskit>=0.4.0
+qiskit>=0.46.0
```
Cynocracy, line 1: should we leave this out so that the user can select the version? I've seen issues with qiskti 1.0 not being downgrade-able so I'm a little hesitant to have a default dependency on it.

Cynocracy (approved):

requirements.txt
splch: we wanted to add a dependencies for `>=0.46.0` and `>=1.0.0` versions but couldnt find a good way to do that with pip. maybe just having `qiskit` be a dependency without a version until `0.46.x` is unsupported would be fine?

Cynocracy: > I ran into this comment while searching for something else. With which version of qiskit is qiskit-ionq compatible with? I don't fully get the advantage of removing the dependency and failing at import time.

Qiskit 1.0 was not compatible with many qiskit ecosystem libs at launch. If we were to leave qiskit without a version in the requirements, then it would by default install qiskit at the latest version, which may be 1.0+, and cause issues for users who want to use pre-qiskit-1.0 libraries.

By removing the dependency entirely and failing on import with a warning, we allow the user to select whether they would like to use qiskit 1.0 or qiskit < 1.0 without requiring them to 'undo' the changes that the install made for them.

Theoretically qiskit without a version dependency would be fine, @splch , but I was worried that it was no longer possible to `pip install qiskit-ionq` and get the same behavior re: support for other libraries from scratch, because 'from scratch' you now default get qiskit 1.0

Anyways, I would be fine with qiskit without a particular version, but I'd like to have docs about how to downgrade from qiskit 1.0 to 0.46 in that case, since it seems like a path that users may have to deal with.

Cynocracy: If you install qiskit 0.45 and qiskit 1.0 into the same python environment, they will fail, because qiskit 0.45 install qiksit-terra 0.45 and this later conflicts with qiskit 1.0. This seems like a fundamentally flawed upgrade path and I am leery to put people on it by default, I suppose.

PR #175, opened 2024-03-21
add support for .env file loading keys

splch:
### Summary

can create a .env file and use project-specific `IONQ_API_KEY`.

### DIFF
--- a/qiskit_ionq/ionq_provider.py
+++ b/qiskit_ionq/ionq_provider.py
@@ -28,6 +28,7 @@
 
 import logging
 import os
+from dotenv import load_dotenv
 
 from qiskit.providers.exceptions import QiskitBackendNotFoundError
 from qiskit.providers.providerutils import filter_backends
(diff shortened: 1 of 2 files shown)


qiskit_ionq/ionq_provider.py
```diff
@@ -28,13 +28,20 @@
...
 
 from . import ionq_backend
 
 logger = logging.getLogger(__name__)
+load_dotenv(
+    dotenv_path=os.path.join(
+        os.getcwd(), ".env"
+    ),  # Load .env file from current directory
+    override=True,  # Set override=True to allow for .env file to override env vars
+)
```
Cynocracy, line 44: It's pretty strange for an SDK to load env files like this on import. Why would you want this?

Cynocracy: In this mode, basically, all packages referenced before qiskti-ionq will have the normal env vars, then when ionq loads, it would potentially change these.

Could we use dotenv_values to avoid mucking with the environment?

splch: yeah that's a lot better thanks for the review!

Cynocracy (approved):

PR #185, opened 2024-07-31
add test for meas_mapped

splch:
### Summary

ensures we dont break qubit mapping again

### DIFF
--- a/test/helpers/test_qiskit_to_ionq.py
+++ b/test/helpers/test_qiskit_to_ionq.py
@@ -206,6 +206,58 @@ def test_full_circuit(simulator_backend):
     assert actual == expected_rest_of_payload
 
 
+def test_multicircuit_mapping(simulator_backend):
+    """Test multiple circuits with mapped qubits
+
+    Args:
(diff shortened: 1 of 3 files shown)


test/helpers/test_qiskit_to_ionq.py
```diff
@@ -206,6 +206,58 @@ def test_full_circuit(simulator_backend):
...
+        "metadata": {
+            "shots": "1024",
+            "sampler_seed": "None",
+        },
+        "noise": {"model": "ideal", "seed": None},
+    }
+
+    actual = json.loads(ionq_json)
+    actual["metadata"].pop("qiskit_header")  # remove qiskit_header for comparison
+    assert actual == expected_payload
```
Cynocracy, line 258: any way to test the flip-side? On a map with particular values, that the results are mapped correctly?

Cynocracy (approved):

test/helpers/test_qiskit_to_ionq.py
splch: oh ya let me add that

PR #186, opened 2024-08-07
Remove qiskit > 1.0.0 requirement

Cynocracy:
As noted in qiskit's docs, it's potentially problematic to upgrade a pre-existing python environment to qiskit 1.0.0.

Given that we don't require any 1.0.0 functionality (i.e, are currently cross-compatible), we can relax the versioning requirement and just require qiskit.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,4 @@ requests>=2.24.0
 retry>=0.9.0
 importlib-metadata>=4.11.4
 python-dotenv>=1.0.1
-qiskit>=1.0.0
+qiskit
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -3,4 +3,4 @@ requests>=2.24.0
 retry>=0.9.0
 importlib-metadata>=4.11.4
 python-dotenv>=1.0.1
-qiskit>=1.0.0
+qiskit
```
Cynocracy, line 6: previously I had removed this entirely because it worked well when installed with either qiskit < 1.0 or > 1.0 "out of the box", will test the same here.

Cynocracy: Requirement already satisfied: qiskit-terra==0.46.2 in /home/jon/.pyenv/versions/qiskit-old/lib/python3.11/site-packages (from qiskit->qiskit-ionq==0.5.4) (0.46.2)

yup seems to work OK

splch (approved):

PR #188, opened 2024-09-04
get num of qubits from api or use 36 and fix EM sampler

splch:
### Summary

temporary solution before migrating to [backendv2](https://docs.quantum.ibm.com/api/qiskit/qiskit.providers.models.BackendConfiguration)

### Details and comments

update `n_qubits` to `36`.
set EM from backend options

### DIFF
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -34,6 +34,8 @@ import gzip
 import base64
 import platform
 import warnings
+import os
+import requests
 
 from qiskit import __version__ as qiskit_terra_version
 from qiskit.circuit import controlledgate as q_cgates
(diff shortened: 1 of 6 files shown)


qiskit_ionq/helpers.py
```diff
@@ -474,6 +477,73 @@ def default(self, o):
...
+    url, token = resolve_credentials().values()
+    try:
+        return requests.get(
+            url=f"{url}/characterizations/backends/{backend}/current",
+            headers={"Authorization": f"apiKey {token}"},
+            timeout=5,
+        ).json()["qubits"]
+    except Exception:  # pylint: disable=broad-except
+        if backend == "ionq_qpu.harmony":
+            return 11
```
Cynocracy, line 536: why this? falling back to a hard-coded value when the API is down feels a little weird

qiskit_ionq/helpers.py
```diff
@@ -474,6 +477,73 @@ def default(self, o):
...
+        elif backend == "ionq_qpu.aria-1":
+            return 25
+        elif backend == "ionq_qpu.aria-2":
+            return 25
+        elif backend == "ionq_qpu.forte-1":
+            return 36
+        else:
+            return 36
```
Cynocracy, line 544: I would have a hard time understanding what this code was doing, since it falls back during an exception but doesn't' log it or what value it's deciding. Could you add logs here? (or alternatively, fall back to Infinity or something >> 36?)

Cynocracy (changes requested):

qiskit_ionq/helpers.py
splch: because the user might not have an api_key set, and we require that to get characterization data (including the number of qubits)

qiskit_ionq/helpers.py
splch: we can't do inifinity because the transpiler freezes (i guess it tries to create a graph of length #qubits lol

Cynocracy (approved):

qiskit_ionq/helpers.py
Cynocracy: ah I meant more, the part where we code in the current system sizes, would worry it could quickly become stale, like the new method with warning and a fixed cap, thanks for changing!

PR #189, opened 2024-09-06
Fix mismatched assignment of URL/token

Cynocracy:


Cynocracy:
In the future, we need to be adding tests for this new functionality

splch (approved):

Cynocracy:
ty!

PR #191, opened 2024-09-09
Add type annotations to part of the public API

basnijholt:

### DIFF
--- a/qiskit_ionq/__init__.py
+++ b/qiskit_ionq/__init__.py
@@ -26,8 +26,6 @@
 
 """Provider for IonQ backends"""
 
-import warnings
-
 # warn if qiskit is not installed
 try:
     from qiskit.version import get_version_info  # pylint: disable=unused-import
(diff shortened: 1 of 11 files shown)


qiskit_ionq/__init__.py
```diff
@@ -26,8 +26,6 @@
 
 """Provider for IonQ backends"""
 
-import warnings
```
basnijholt, line 29: Was unused.

Cynocracy (approved):

Cynocracy:
Looks great, thanks for adding @basnijholt !

setup.py
```diff
@@ -78,7 +78,7 @@
         "Topic :: Scientific/Engineering",
     ],
     keywords="qiskit sdk quantum",
-    python_requires=">=3.7",
+    python_requires=">=3.9",
```
basnijholt, line 81: Python 3.7 is already EOL and 3.8 will be at the end of this month.

Cynocracy: makes sense, ty again!

Issue #190, opened 2024-09-06
Dynamically loading qubit count per backend


Cynocracy:
Functionality added in qiskit-ionq 0.5.5 to dynamically load the backend size can leak sensitive info to logs, and is being pulled.

Cynocracy:
@splch we can keep this one around until 0.5.6 is ready for release if that works for you?

PR #196, opened 2024-09-12
Trim ionq_ prefix from backends lookup

Cynocracy:
@splch do you think we could target having tests added for this before releasing 0.5.6? I'm a little uncomfortable with it given the hiccups re: launch

https://github.com/qiskit-community/qiskit-ionq/issues/190


splch (approved):
great with tests working this is good to go :)

PR #200, opened 2024-10-14
allow setting multicircuit names from run method

splch:
### Summary

can do:

```python
multi_result = backend.run([qc_1, qc_2], name="multi-circuit").result().get_counts()
```

### Details and comments

when running many jobs, concatenating child circuit names wont scale. this lets you set a short name

### DIFF
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -411,7 +411,7 @@ def qiskit_to_ionq(
     ionq_json = {
         "target": target,
         "shots": passed_args.get("shots"),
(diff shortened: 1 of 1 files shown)


Cynocracy (changes requested):
Can you add a test case demonstrating what you mean? This seems to allow you to override every circuit name in the passed arguments, is that what you want?

splch:
> Can you add a test case demonstrating what you mean? This seems to allow you to override every circuit name in the passed arguments, is that what you want?

yeah i shouldve made this a draft pr - ill add some tests once it's working :)

Cynocracy (approved):

PR #202, opened 2024-10-21
fix default qubit warning for simulator

splch:
### Summary

the simulator doesn't have accessible qubit defaults, so we need to manually add this

### Details and comments

prevents a warning from the backend initialization

### DIFF
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -38,6 +38,7 @@ import platform
 import warnings
 import os
 from typing import Literal, Any
+from functools import lru_cache
 import requests
 from dotenv import dotenv_values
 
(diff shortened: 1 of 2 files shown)


gmauricio (approved):

qiskit_ionq/helpers.py
```diff
@@ -535,35 +539,43 @@ def resolve_credentials(token: str | None = None, url: str | None = None):
...
 
     Args:
         backend (str): The name of the backend.
+        fallback (int): Fallback number of qubits if API call fails.
 
     Returns:
         int: The number of qubits for the backend.
     """
+    # Handle known case for the simulator directly
+    if backend == "simulator":
```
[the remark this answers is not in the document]
splch: i just updated it to use our publicly accessible backends endpoint which provides qubit counts :D

qiskit_ionq/helpers.py
```diff
@@ -515,17 +517,19 @@ def resolve_credentials(token: str | None = None, url: str | None = None):
     Returns:
         dict[str]: A dict with "token" and "url" keys, for use by a client.
     """
+    # Cache dotenv and environment variable lookups
+    env_values = dotenv_values()
```
basnijholt, line 521: This will mean that if one calls `resolve_credentials` and doesn't have the env vars set, then sets it, then calls `resolve_credentials` again, which will return the incorrect thing.

No need to cache such a cheap operation IMHO.

splch: good point - ill remove caching

PR #203, opened 2024-10-23
Fix bug regarding serializing custom exceptions which occurs in multiprocessing context

basnijholt:
### Summary

The bug can be reproduced with:
```python
from qiskit_ionq.exceptions import IonQAPIError
import pickle
err = IonQAPIError(message='hi', status_code=400, headers='hi', body='hi', error_type='hi')
pickle.loads(pickle.dumps(err))
```
which gives:
```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[5], line 4
      2 import pickle
      3 err = IonQAPIError(message='hi', status_code=400, headers='hi', body='hi', error_type='hi')
----> 4 pickle.loads(pickle.dumps(err))

TypeError: IonQAPIError.__init__() missing 4 required positional arguments: 'status_code', 'headers', 'body', and 'error_type'
```

### DIFF
--- a/qiskit_ionq/exceptions.py
+++ b/qiskit_ionq/exceptions.py
@@ -168,6 +168,18 @@ class IonQAPIError(IonQError):
             f"error_type={self.error_type!r})"
         )
 
+    def __getstate__(self) -> object:
+        return (
+            self.message,
+            self.status_code,
+            self.headers,
+            self.body,
+            self.error_type,
+        )
+
+    def __setstate__(self, state: tuple[object, ...]) -> None:
+        self.message, self.status_code, self.headers, self.body, self.error_type = state
+
 
 class IonQBackendError(IonQError):
     """Errors generated from improper usage of IonQBackend objects."""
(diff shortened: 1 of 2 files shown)


amritpoudelionq:
@basnijholt also specify OS?

Cynocracy (approved):

basnijholt:
@amritpoudelionq, the problem should occur on any OS.

qiskit_ionq/exceptions.py
```diff
@@ -168,18 +168,12 @@ def __str__(self):
             f"error_type={self.error_type!r})"
         )
 
-    def __getstate__(self) -> object:
+    def __reduce__(self):
```
amritpoudelionq, line 171: do we understand why __reduce__ works? __get__ and __set__ are pretty standard way for serialization?

splch: i just tested this out and it seems like `__getstate__` and `__setstate__` are relying on an instantiated object (which is why we see the error `TypeError: IonQAPIError.__init__() missing 4 required positional arguments: 'status_code', 'headers', 'body', and 'error_type'`) but `__reduce__` returns the class and arguments to instantiate it.

PR #205, opened 2024-12-02
remove retry dependency

splch:
### Summary

removes the retry dependency because of the py vulnerability. adds a custom retry decorator that does the same thing without external deps

### Details and comments

https://github.com/invl/retry/issues/58

### DIFF
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -576,7 +576,8 @@ def retry(
     max_delay: float = float("inf"),
     backoff: float = 1,
     jitter: float = 0,
-):  # pylint: disable=too-many-positional-arguments
+    logging: bool = True,
+) -> function:  # pylint: disable=too-many-positional-arguments
     """Retry decorator with exponential backoff.
 
     Args:
@@ -586,6 +587,7 @@ def retry(
         max_delay: Maximum delay between retries.
         backoff: Multiplier applied to delay after each retry.
         jitter: Maximum random jitter added to delay.
+        logging: Whether to log failures.
     """
 
     def deco_retry(func):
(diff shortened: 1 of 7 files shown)


qiskit_ionq/helpers.py
```diff
@@ -566,6 +569,54 @@ def get_n_qubits(backend: str, _fallback=100) -> int:
...
+        tries (int): number of times to try (not retry) before giving up. -1 means infinite tries.
+        delay (float): initial delay between retries in seconds.
+        max_delay (float): the maximum value of delay in seconds. default is no max.
+        backoff (float): backoff multiplier e.g. value of 2 will double the delay each retry.
+        jitter (float or callable): extra seconds added to delay between retries to
+            prevent thundering herd effect. may be fixed value or a callable that
+            returns a value.
+    """
+
+    def deco_retry(func):
```
Cynocracy, line 594: If we're going to write our own, can we add tests? Separately, log statements on failures would also be useful. 

https://pypi.org/project/reretry/ came up as another alternative as well

Cynocracy (changes requested):

qiskit_ionq/helpers.py
splch: added a test and logging :)

test/helpers/test_helpers.py
```diff
@@ -115,3 +115,41 @@ def test_get_n_qubits_fallback():
...
+
+    result = func_success()
+    assert (
+        attempt_success["count"] == 2
+    ), f"Expected 2 retries, got {attempt_success['count']}"
+    assert result == "Success", f"Expected 'Success', got {result}"
+
+    # Test case where the function keeps failing and eventually raises the exception
+    attempt_fail = {"count": 0}
+
```
Cynocracy, line 140: One more for exception of the wrong type? Thx for adding!

Cynocracy (approved):

PR #214, opened 2024-12-10
catch all multi target uncontrolled gates

splch:
### Summary

instead of checking against a tuple, check the quality of the gate

### Details and comments

`if the gate has multiple targets and no control qubits`

### DIFF
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -238,7 +238,8 @@ def qiskit_circ_to_ionq_circ(
         # Make sure uncontrolled multi-targets use all qargs.
         if instruction.num_qubits > 1 and not hasattr(instruction, "num_ctrl_qubits"):
             converted["targets"] = [
-                input_circuit.qubits.index(qargs[i]) for i in range(instruction.num_qubits)
+                input_circuit.qubits.index(qargs[i])
+                for i in range(instruction.num_qubits)
             ]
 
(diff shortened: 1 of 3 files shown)


qiskit_ionq/helpers.py
```diff
@@ -244,7 +236,7 @@ def qiskit_circ_to_ionq_circ(
             converted["gate"] = instruction_name
 
         # Make sure uncontrolled multi-targets use all qargs.
-        if isinstance(instruction, multi_target_uncontrolled_gates):
+        if instruction.num_qubits > 1 and not hasattr(instruction, "num_ctrl_qubits"):
```
Cynocracy, line 239: is num_qubits guaranteed to be set?

I don't quite understand why we are changing this and not https://github.com/qiskit-community/qiskit-ionq/blob/08a15dc4ae19880533d0732d9eab14766fe9341a/qiskit_ionq/helpers.py#L193-L194 based on the gates you've seen issues with?

Cynocracy: Spoke offline: main issue is that where I linked we check gate name, and here we are checking python classes, which might not match in the case of QASM/etc

WDYT about changing to the values from GATESET_MAP that we know should have multiple controls?

This LGTM assuming the num_qubits attribute must exist, if it might not I would change to getattr(instruction, "num_qubits", 0)

Cynocracy (approved):

PR #234, opened 2025-09-02
Add memory support to IonQJob and update tests for error handling

splch:
### Summary

adds memory to support shotwise output

### Details and comments

- need to get shotwise endpoint only when requested
- do not attempt shotwise for ideal simulations

### DIFF
--- a/qiskit_ionq/ionq_job.py
+++ b/qiskit_ionq/ionq_job.py
@@ -329,7 +329,7 @@ class IonQJob(JobV1):
         if self._status is jobstatus.JobStatus.DONE:
             assert self._job_id is not None
             response = self._client.get_results(
-                results_url=self._results_url,
+                results_url=self._results_urls.get("probabilities", ""),
                 sharpen=sharpen,
                 extra_query_params=extra_query_params,
             )
@@ -391,14 +391,12 @@ class IonQJob(JobV1):
                 self._num_circuits = self._first_of(stats, "circuits", default=1)
 
             self._num_qubits = self._first_of(stats, "qubits", default=0)
-            _results_url = self._first_of(
-                response, "results", "results_url", default={}
-            )
-            self._results_url = (
-                _results_url
-                if isinstance(_results_url, str)
-                else _results_url.get("probabilities", {}).get("url")
-            )
+            results = response.get("results", {})
+            self._results_urls = {
+                k: v.get("url")
+                for k, v in results.items()
+                if isinstance(v, dict) and "url" in v
+            }
 
             # Classical-bit maps per circuit
             def _meas_map_from_header(header_dict, fallback_nq):
--- a/qiskit_ionq/ionq_job.py
+++ b/qiskit_ionq/ionq_job.py
@@ -46,6 +46,7 @@ from qiskit.providers import JobV1, jobstatus
 from qiskit.providers.exceptions import JobTimeoutError
 from .ionq_result import IonQResult as Result
 from .helpers import decompress_metadata_string, normalize
+from .exceptions import IonQBackendError
 
 from . import constants, exceptions
 
@@ -180,10 +181,12 @@ class IonQJob(JobV1):
         if passed_args is not None:
             self.extra_query_params = passed_args.pop("extra_query_params", {})
             self.extra_metadata = passed_args.pop("extra_metadata", {})
+            self.memory = passed_args.pop("memory", True)
             self._passed_args = passed_args
         else:
             self.extra_query_params = {}
             self.extra_metadata = {}
+            self.memory = True
             self._passed_args = {"shots": 1024, "sampler_seed": None}
 
         # Support single or list-of-circuits submissions
@@ -243,6 +246,23 @@ class IonQJob(JobV1):
             dict: A dictionary of counts.
         """
         return self.result().get_counts(circuit)
+
+    def get_memory(self, circuit=None):
+        """
+        Return the memory for the job.
+        Args:
+            circuit (str or QuantumCircuit or int or None): Optional.
+        Returns:
+            list: A list of memory strings.
+        """
+        if self.memory:
+            return self.result().get_memory(circuit)
+
+        raise IonQBackendError(
+            f'No memory for experiment "{circuit.name if circuit else ""}". '
+            "Please verify that you ran a job with "
+            'the memory flag set, eg., "memory=True".'
+        )
 
     def get_probabilities(self, circuit=None):  # pylint: disable=unused-argument
         """Return the probabilities (for simulators).
@@ -551,8 +571,10 @@ class IonQJob(JobV1):
                     use_sampler=is_ideal_sim,
                     sampler_seed=sampler_seed,
                 )
+                memory = None
                 job_result[i]["data"] = {
                     "counts": counts,
+                    "memory": memory,
                     "probabilities": probabilities,
                     "metadata": qiskit_header[i] or {},
                 }
--- a/qiskit_ionq/ionq_job.py
+++ b/qiskit_ionq/ionq_job.py
@@ -577,7 +577,8 @@ class IonQJob(JobV1):
 
         if isinstance(data, dict):
             looks_like_multi = all(
-                isinstance(v, dict) and all(isinstance(p, (int, float)) for p in v.values())
+                isinstance(v, dict)
+                and all(isinstance(p, (int, float)) for p in v.values())
                 for v in data.values()
             )
             data = list(data.values()) if looks_like_multi else [data]
--- a/qiskit_ionq/ionq_job.py
+++ b/qiskit_ionq/ionq_job.py
@@ -38,7 +38,7 @@
 from __future__ import annotations
 
 import warnings
-from typing import TYPE_CHECKING, Any, Optional, Callable
+from typing import TYPE_CHECKING, Any, Callable
 import numpy as np
 
 from qiskit import QuantumCircuit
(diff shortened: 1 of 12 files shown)


splch (commented):
comment below vv

qiskit_ionq/ionq_job.py
```diff
@@ -543,16 +602,31 @@ def _format_result(self, data):
...
+                    memory = _build_memory(
+                        raw_shots=self._client.get_results(
+                            f"/v0.4/jobs/{job_id}/results/shots"
+                        ),
+                        n_qubits=qiskit_header[i].get("n_qubits", self._num_qubits),
+                        clbits=self._clbits[i],
+                        width=(qiskit_header[i] or {}).get(
+                            "memory_slots", len(self._clbits[i])
+                        ),
+                    )
```
splch: i dont like how this looks

splch:
We standardized cross-SDK on `memory` + off by default (this PR ✅). One ask before merge: add multi-circuit support. Each child job advertises its own `results.shots.url` — iterate `child_job_ids` and assemble a per-circuit `memory` list. Same pattern as PennyLane-IonQ#178's `_retrieve_child_job_shots`. This unblocks #237 properly instead of forcing it down the resample path.

qiskit_ionq/ionq_job.py
```diff
@@ -144,6 +145,33 @@ def _build_counts(  # pylint: disable=too-many-positional-arguments
...
+    width: int | None = None,
+) -> list[str]:
+    """Convert IonQ shot outcomes into Qiskit memory bitstrings.
+
+    The IonQ v0.4 ``CircuitJobResult.shots.url`` endpoint serves an ordered
+    list of decimal-encoded outcome integers (same convention as histogram
+    keys); this function applies the ``clbits`` mapping and returns one
+    MSB-left bitstring per shot.
+    """
+    if not clbits:
```
guenp:
```suggestion
    if clbits is None:
```

I would make this an explicit check, following the Zen of Python

qiskit_ionq/ionq_job.py
```diff
@@ -144,6 +145,33 @@ def _build_counts(  # pylint: disable=too-many-positional-arguments
...
+        clbits = list(range(n_qubits))
+    out_width = int(width) if width is not None else len(clbits) if clbits else n_qubits
+
+    def remap_one(val: int | str) -> str:
```
guenp:
```suggestion
    def remap_raw_shot(val: int | str) -> str:
```
naming nit

qiskit_ionq/ionq_job.py
```diff
@@ -144,6 +145,33 @@ def _build_counts(  # pylint: disable=too-many-positional-arguments
...
+        raw = bin(int(val))[2:].rjust(n_qubits, "0")[::-1]
+        mapped = "".join(
+            raw[b] if b is not None and 0 <= b < n_qubits else "0" for b in clbits
+        )[::-1]
+        return mapped.rjust(out_width, "0")
+
+    return [remap_one(s) for s in raw_shots]
```
guenp:
```suggestion
    return [remap_raw_shot(s) for s in raw_shots]
```

qiskit_ionq/ionq_job.py
```diff
@@ -291,6 +321,22 @@ def get_counts(self, circuit: Optional[QuantumCircuit] = None) -> dict:
...
         return self.result().get_counts(circuit)
 
+    def get_memory(self, circuit=None):
+        """Return per-shot memory bitstrings (MSB-left), one entry per shot.
+
+        ``circuit`` follows the same semantics as :meth:`get_counts`.
+        Raises :class:`IonQBackendError` if the job was submitted with
+        ``memory=False``.
+        """
+        if self.memory:
```
guenp, line 333:
```suggestion
        if self.memory is True:
```

just better to be explicit imo (for code readability)

qiskit_ionq/ionq_job.py
```diff
@@ -430,18 +476,14 @@ def status(self, detailed: bool = False) -> jobstatus.JobStatus | dict:
...
-                _results_url = self._first_of(
-                    response, "results", "results_url", default={}
-                )
-                self._results_url = (
-                    _results_url
-                    if isinstance(_results_url, str)
-                    else _results_url.get("probabilities", {}).get("url")
-                )
+                results = response.get("results") or {}
+                self._results_urls = {
```
guenp, line 484: Does this satisfy the multi-circuit support mentioned in the most recent comment?

qiskit_ionq/ionq_job.py
guenp: the docstring doesn't match the function signature, and could use more clarity (i.e. what is the expected schema of `result`)

qiskit_ionq/ionq_job.py
```diff
@@ -587,7 +648,8 @@ def _format_result(self, data):
 
         if isinstance(data, dict):
             looks_like_multi = all(
-                isinstance(v, dict) and all(isinstance(p, float) for p in v.values())
+                isinstance(v, dict)
+                and all(isinstance(p, (int, float)) for p in v.values())
```
guenp, line 696: I'll need a bit more context for this change. Am I inferring correctly that the output format of `data` changed? It looks like it went from `{str: float}` to `{str: (int, float)}`? that seems like a breaking change, and would perhaps require a version bump (if so please note in the PR description). On a related note, should we also check for the old format here for backward compatibility?

qiskit_ionq/ionq_job.py
```diff
@@ -610,26 +672,51 @@ def _format_result(self, data):
...
-                    qiskit_header[i].get("n_qubits", len(clbits) or self._num_qubits),
+                    n_qubits,
                     clbits,
                     shots,
                     use_sampler=is_ideal_sim,
                     sampler_seed=sampler_seed,
                 )
+                if raw_shots is not None:
+                    memory_slots = header.get("memory_slots") or len(clbits or [])
+                    memory = _build_memory(raw_shots, n_qubits, clbits, memory_slots)
```
guenp, line 751:
```suggestion
                    memory = _build_memory(raw_shots=raw_shots, n_qubits=n_qubits, clbits=clbits, width=memory_slots)
```

It can be a little risky to skip the variable names, especially if the variable names and the function input names don't match, I would recommend making this explicit

```
_build_memory(
    raw_shots: list[int | str],
    n_qubits: int,
    clbits: list[int] | None,
    width: int | None = None,
)
```

qiskit_ionq/ionq_job.py
```diff
@@ -144,6 +145,33 @@ def _build_counts(  # pylint: disable=too-many-positional-arguments
     return counts, probabilities
 
 
+def _build_memory(
```
guenp, line 148:
```suggestion
def _shots_to_memory_bitstrings(
```
nit: more readable/understandable naming
I think "build" is not really accurate here - we are _converting_ one encoding to another, so imo `a_to_b` would be a better naming paradigm to use

guenp (approved):
lgtm, a few naming and value assertion nits, and curious about the multi-circuit support in the last comment -- it says "before merging", does that mean this is a fix to still be added?

qiskit_ionq/ionq_job.py
splch: Done — switched to `if clbits is None:` so the check lines up with the `list[int] | None` signature.

qiskit_ionq/ionq_job.py
splch: Done.

qiskit_ionq/ionq_job.py
splch: Done (paired with the helper rename above).

qiskit_ionq/ionq_job.py
splch: Going to keep `if self.memory:` here. PEP 8 *Programming Recommendations* is explicit about this one:

> Don't compare boolean values to True or False using `==`. Use `if greeting:` instead of `if greeting is True:`.

`self.memory` is initialized as a `bool` in `__init__` (`passed_args.pop("memory", False)`), so the truthy form is the idiomatic Python check. The identity comparison would also silently break if someone ever flowed a truthy non-`True` value through that field.

qiskit_ionq/ionq_job.py
splch: No — that block just records `shots` and `probabilities` as separate URL slots so the rest of the code can pick one. The actual multi-circuit work is in the new commit on top.

`_raw_shots_per_circuit` now iterates `child_job_ids` and fetches each child's own `results.shots.url` (the parent only advertises aggregated probabilities for multi-circuit jobs). Failures degrade per-circuit — one bad child returns `memory=None` for that slot, the others still assemble. Covered by `test_multi_mem_per_child` (happy path) and `test_multi_mem_partial` (one child missing its shots URL).

qiskit_ionq/ionq_job.py
splch: Rewrote it: the param is `data` (not `result`), and the docstring now spells out the three accepted shapes (`dict[str, float]` single, `dict[str, dict[str, float]]` multi keyed by child id, and the `list[dict[str, float]]` tests use) plus the `int`/`float` accepted for probability values.

qiskit_ionq/ionq_job.py
splch: Not a format change — strictly a permissiveness bump. Python's `json` module deserializes the JSON literals `0` and `1` as Python `int`, not `float`, so a noiseless-simulator child whose probabilities happen to be exact 0/1 used to fail the heuristic, and `_format_result` silently mis-classified that multi-circuit payload as single-circuit. Old `{str: float}` payloads still match — adding `int` to the isinstance tuple makes the check more permissive, not less. No version bump and no back-compat shim needed; the same `data: dict[str, float]` callers continue to work. The new `_format_result` docstring calls this out so the schema is explicit going forward.

qiskit_ionq/ionq_job.py
splch: Fixed at the root rather than at the call site: renamed the parameter `width` → `memory_slots` in `_build_memory` so all four positionals match the caller's vocabulary directly. Sprinkling kwargs at the call site would have preserved the impedance mismatch (`memory_slots` → `width`) that prompted the comment; aligning the names is the cleaner read.

qiskit_ionq/ionq_job.py
splch: Going to keep `_build_memory`. It has a sibling 30 lines above — `_build_counts` — and the two are the producers `_format_result` assigns into `data.counts` and `data.memory`. Renaming this one alone breaks the established naming family; renaming both is out of scope here (and `_build_counts` does more than convert — it samples from probabilities for the ideal simulator, so "build" actually fits it).

`memory` is also the Qiskit term of art for the list of per-shot bitstrings that lives at `result.data.memory`, so `_build_memory` reads as "build the `memory` field," not "build memory in the OS sense."

splch:
@guenp thanks for the careful pass. Two new commits address the round.

**Multi-circuit memory** (the load-bearing pre-merge ask): `_raw_shots_per_circuit` iterates `child_job_ids` and fetches each child's own `results.shots.url` — the parent only advertises aggregated probabilities for multi-circuit, so the previous "decline rather than mis-assign" branch was leaving real per-shot data on the table. Failures degrade per-circuit; one bad child returns `memory=None` for that slot without poisoning the rest. Covered by `test_multi_mem_per_child` and `test_multi_mem_partial`.

**Review nits accepted:** `if clbits is None` (matches the `list[int] | None` signature), `remap_one` → `remap_raw_shot`, `_format_result` docstring rewritten with the three accepted payload shapes and the `int`/`float` probability values. Also addressed the `_build_memory` call-site mismatch by renaming the parameter `width` → `memory_slots` so all four positionals match the caller's vocabulary directly (cleaner than spreading kwargs at the call site).

**Two declined,** with rationale in the inline threads:
- `if self.memory is True:` — PEP 8 explicitly forbids this form for boolean checks.
- `_build_memory` → `_shots_to_memory_bitstrings` — would break the naming pair with the existing `_build_counts` sibling.

The `(int, float)` change is a permissiveness bump (JSON `0`/`1` deserialize as Python `int`), not a format change — old `{str: float}` payloads still match. No version bump needed; the new docstring documents the schema explicitly.

Full suite (392 tests) green on `pr-234`; pylint clean on both `qiskit_ionq/` and `test/`.

PR #237, opened 2025-12-05
add IonQSampler and IonQEstimator classes

splch:
### Summary

adds support for the sampler and estimator primitives. resolves issue #242 

### Details and comments

the sampler expects `memory=True`, so this will need to be updated after #234 is merged
also added testing for python 3.14 🐍🥧

### DIFF
--- a/qiskit_ionq/ionq_primitives.py
+++ b/qiskit_ionq/ionq_primitives.py
@@ -41,8 +41,11 @@ from qiskit.primitives.containers import (
     PrimitiveResult,
     SamplerPubLike,
     SamplerPubResult,
+    EstimatorPubLike,
+    EstimatorPubResult,
 )
 from qiskit.primitives.containers.sampler_pub import SamplerPub
+from qiskit.primitives.containers.estimator_pub import EstimatorPub
 from qiskit.primitives.primitive_job import PrimitiveJob
 
 from .ionq_backend import IonQBackend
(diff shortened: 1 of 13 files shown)


qiskit_ionq/ionq_primitives.py
```diff
@@ -0,0 +1,164 @@
...
+                  (circuit, parameter_values, [shots]) tuples.
+            shots: Default shots for pubs that don't specify their own.
+        """
+        if shots is None:
+            shots = self._default_shots
+
+        # Normalize everything to SamplerPub objects with shots filled in.
+        coerced_pubs = [SamplerPub.coerce(pub, shots) for pub in pubs]
+
+        job = PrimitiveJob(self._run, coerced_pubs)
```
hdnq, line 116: @splch I think this is the origin of one of the user-reported bugs, which is not fixed here - qiskit's BackendSamplerV2 creates and returns a PrimitiveJob (with random uuid) to the user, and the same thing is still happening here. There's an IonQJob created in `_run_pub` but only the result, not the job, gets passed back up.

so when a user does this
```
job = sampler.run([qc], shots=100)
job.job_id()
```
they'll still just get a random uuid back and not the uuid for the actual job in our platform. They want the actual uuid for the job so they can store it, retrieve the result later, etc. (I think Estimator does the same thing)

splch: thanks for the reminder! i'll get that added

Cynocracy (approved):

PR #243, opened 2026-04-17
Removing old client code.

radumarg:


radumarg:
The ionq-core API is not yet ready

PR #245, opened 2026-04-30
Adopt ionq-core as the HTTP transport, gate-matrix source, and exception bridge

splch:
## Summary

Add [`ionq-core`](https://github.com/ionq/ionq-core-python) as a runtime dependency and route every outbound HTTP call, the four native-gate unitaries, and the API-error translator through it. The `IonQClient` facade is preserved (every method now delegates to a typed `ionq-core` endpoint) so callers — `IonQJob`, `IonQBackend`, `Session` — don't change in this PR. The legacy public surface (`IonQClient` method names + signatures, `IonQAPIError` / `IonQRetriableError` hierarchy, `Characterization` / `JobEstimate` wrappers, `ErrorMitigation` enum) is unchanged.

This is the **first step** in a stack that will collapse `qiskit_ionq/main` onto the [`ionq-core` end-state branch](https://github.com/qiskit-community/qiskit-ionq/tree/ionq-core). Each subsequent PR is monotonically subtractive — adds one boundary translator (`_translate.py` / `_submit.py` / `_result.py` / `_target.py`) from the end state and deletes the corresponding piece of `helpers.py`/`exceptions.py`/etc. that the end state lacks.

## What changes

- **`requirements.txt`** — `+ ionq-core>=0.1.0`. Pinned to the PyPI release; no git-pin and no `do-not-merge` block.
- **`setup.py` / `tox.ini` / `.github/workflows/ci.yml`** — drop Python 3.10/3.11 (ionq-core requires ≥3.11; this repo's end state is ≥3.12), add 3.14.
- **`qiskit_ionq/ionq_gates.py`** — `GPIGate`/`GPI2Gate`/`MSGate`/`ZZGate` `__array__` now calls `ionq_core.{gpi,gpi2,ms,zz}_matrix` instead of recomputing the closed-form unitaries by hand. Single source of truth for IonQ-canonical matrices. (-48 lines.)
- **`qiskit_ionq/exceptions.py`** — `IonQAPIError.from_ionq_core(exc)` classmethod translates `ionq_core.exceptions.APIError` (and its typed subclasses) into the existing `IonQAPIError` shape, preserving the IonQ Cloud `request_id` for support escalation. Marked `TODO(ionq-core): transitional`; deleted in a follow-up PR alongside the rest of `qiskit_ionq.exceptions`.
- **`qiskit_ionq/ionq_client.py`** — every method (`submit_job`, `retrieve_job`, `cancel_job`, `delete_job`, `get_results`, `estimate_job`, `get_calibration_data`, `post`, `put`, `get_with_retry`) now dispatches through the corresponding typed `ionq_core.api.*` endpoint:
  - `create_job` / `get_job` / `cancel_job` / `delete_job` / `get_job_probabilities` / `estimate_job_cost` for the job lifecycle
  - `get_characterizations_for_backend` for calibration
  - `create_session` / `end_session` for `Session.{create, end}`
  - generic POST/PUT helpers raise `IonQClientError` for arbitrary URLs (every supported call has a typed endpoint)
- **`qiskit_ionq/helpers.py`** — `get_n_qubits` now reads `Backend.qubits` via `ionq_core.api.backends.get_backend`. `@lru_cache(maxsize=None)` caches per process (qubit counts don't change at runtime); `max_retries=0` keeps unauthenticated introspection cheap when the API is unreachable.
- **Test infrastructure**:
  - `requirements-test.txt`: `requests-mock` → `pytest-httpx` (the wire is now `httpx`-backed).
  - `test/conftest.py` adds a thin shim that maps the legacy `requests_mock` fixture API onto `pytest-httpx` (`.get(url, json=...)`, `.post(...)`, `.request_history[*].json()`). Existing test files do not need per-test edits. The shim is marked transitional; follow-up PRs migrate tests to import-site `@patch` of `ionq-core` symbols (the end-state pattern).
  - Autouse `_httpx_mock_autouse` activates `httpx_mock` for every test and clears the `get_n_qubits` cache so each test sees a clean slate.
  - `request_history` filters out the unauthenticated `GET /backends/<name>` traffic that every backend now issues at construction time, so legacy assertions on history length stay valid.
  - `test/test_mock.py` rewritten against `httpx` (it was a meta-test for the mock library).
  - `test/helpers/test_helpers.py`: two `get_n_qubits` tests previously patched `requests.get` directly; they now use `httpx_mock`.

## Transitional code (flagged for removal in follow-up PRs)

Three pieces of bridging code exist only because this PR keeps the legacy `IonQClient` facade alive:

1. **`IonQAPIError.from_ionq_core` translator** in `qiskit_ionq/exceptions.py` — exists so `except IonQAPIError` keeps catching while callers haven't yet migrated to `ionq_core.exceptions.APIError`. Deleted in the PR that drops the rest of `qiskit_ionq.exceptions`.
2. **`_call_through_core` + `_BareResponse`** in `qiskit_ionq/ionq_client.py` — `ionq-core` 0.1.x's OpenAPI-generated `from_dict` raises `KeyError` on partial mock fixtures used by the test suite. Production responses always include every required field, so the fallback only matters for tests. Deleted once test fixtures migrate to typed `ionq-core` models, or once `ionq-core` ships a lenient parser mode.
3. **`_RequestsMockShim` / `_HistoryItem`** in `test/conftest.py` — preserves the legacy `requests_mock` fixture API. Replaced by import-site `@patch` of `ionq-core` symbols when each test file is rewritten in the end-state pattern.

All three are clearly commented as transitional in the source.

## Path to the end state

After this PR lands, the planned follow-up stack moves `main` toward [`origin/ionq-core`](https://github.com/qiskit-community/qiskit-ionq/tree/ionq-core) one PR at a time. Each PR adds end-state code and deletes legacy code in the same diff:

| PR | Adds | Deletes |
|---|---|---|
| **#246 — translate + submit** | `qiskit_ionq/_translate.py`, `qiskit_ionq/_submit.py` | `helpers.qiskit_to_ionq`, `helpers.qiskit_circ_to_ionq_circ`, `IonQMidCircuitMeasurementError`, `IonQPauliExponentialError`, `IonQGateError`, barrier-strip logic |
| **#247 — result** | `qiskit_ionq/_result.py` | `IonQJob`'s ad-hoc result building, `helpers`'s probability handling |
| **#248 — target + characterization** | `qiskit_ionq/_target.py` | `Characterization` wrapper, `JobEstimate` wrapper, ad-hoc Target construction |
| **#249 — V2 primitives** (rebases #237) | `IonQSampler`, `IonQSamplerJob`, `IonQSession.sampler()` | nothing yet |
| **#250 — error mitigation** | `error_mitigation: dict \| None` callsite shape | `ErrorMitigation` enum |
| **#251 — transpiler plugin** | nothing | `IonQTranspileLevelWarning`, `TrappedIonOptimizerPlugin`, `add_equivalences()`, all the `CancelGPI2Adjoint`/`GPI2TwiceIsGPI`/etc. passes (or extract to a separate package) |
| **#252 — drop legacy exceptions** | nothing | `qiskit_ionq/exceptions.py`, `IonQAPIError.from_ionq_core`, `IonQRetriableError`, `IonQClientError` |
| **#253 — drop legacy backend kwargs** | nothing | `extra_query_params`, `extra_metadata`, `name`, `job_settings`, `ionq_compiler_synthesis` |
| **#254 — tooling switch** | `pyproject.toml` (hatchling, dependency-groups), `uv.lock`, ruff + ty configs, 95% coverage gate | `setup.py`, `tox.ini`, `requirements*.txt`, pylint config |
| **#255 — file renames + final cleanup** | `gates.py`, `provider.py`, `backend.py`, `job.py`, `session.py`, `_IonQGate` base class, README rewrite | `ionq_gates.py`, `ionq_provider.py`, `ionq_backend.py`, `ionq_job.py`, `ionq_session.py`, leftover `helpers.py` |

After #255, `git diff main..ionq-core` is empty. The same playbook (one boundary translator per PR, subtractive net diff) applies to `cirq-ionq` and `pennylane-ionq`.

## Test plan

- [x] `pytest test/` — **337 passed in 3.9s** (Python 3.13, `ionq-core==0.1.1`, `pytest-httpx==0.36.2`)
- [x] Smoke-tested gate matrix parity on a circuit equivalence test
- [x] Smoke-tested `IonQAPIError.from_ionq_core` for status / message / `request_id` propagation
- [ ] CI green on the PR


.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.12", "3.13", "3.14"]
```
guenp, line 15: We need to support 3.11 at least unfortunately, this is the Python version used in the cuQuantum docker container, see https://catalog.ngc.nvidia.com/orgs/nvidia/containers/cuquantum-appliance

qiskit_ionq/exceptions.py
```diff
@@ -160,6 +166,52 @@ def from_response(cls, response: requests.Response) -> IonQAPIError:
...
+
+        ``ionq-core`` raises a typed exception hierarchy keyed on HTTP status
+        (``BadRequestError``/``AuthenticationError``/``NotFoundError``/
+        ``RateLimitError``/``ServerError``/etc.). Code paths that route HTTP
+        through ``ionq-core`` can call this classmethod to surface those
+        failures as an ``IonQAPIError`` with the same ``message``/``status_code``/
+        ``headers``/``body``/``error_type`` shape every other ``qiskit-ionq``
+        call already produces - so ``except IonQAPIError`` keeps catching.
+
+        TODO(ionq-core): transitional bridge. Remove when callers migrate to
```
guenp, line 181: Do we want to keep this todo? How to track them?

qiskit_ionq/exceptions.py
```diff
@@ -160,6 +166,52 @@ def from_response(cls, response: requests.Response) -> IonQAPIError:
             error_type = error_data.get("type") or error_type
         return cls(message, status_code, headers, body, error_type)
 
+    @classmethod
+    def from_ionq_core(cls, exc) -> IonQAPIError:
```
guenp, line 170: Nit: `from_api_error` feels like a better name, since the import could be from elsewhere (another package might in theory import it from ionq_core)

qiskit_ionq/helpers.py
```diff
@@ -639,25 +638,52 @@ def resolve_credentials(token: str | None = None, url: str | None = None) -> dic
...
+
+    Reaches the IonQ API through ``ionq-core``'s typed
+    ``GET /backends/<name>`` endpoint. Returns ``fallback`` on any error
+    (transport failure, missing key, simulator stub, etc.). Results are
+    memoized for the lifetime of the process - backend qubit counts do
+    not change at runtime, and the cache keeps repeated provider/backend
+    constructions from issuing redundant API calls (which also dominates
+    test-suite latency when no real key is configured).
+    """
+    # Local import: avoid circular import at module load time and keep this
```
guenp, line 653: Where is the circular import happening if you import this at the module level?

qiskit_ionq/ionq_client.py
```diff
@@ -24,28 +24,144 @@
...
+    keys eagerly, so any test fixture that builds a partial mock job/results
+    payload (the qiskit-ionq suite has many of these) raises ``KeyError`` from
+    inside the parser. Production responses from ``api.ionq.co`` always
+    include every required field, so this fallback only matters for tests.
+    When parsing fails we re-issue the request via the underlying httpx
+    client and return a synthetic ``Response``-shaped object whose
+    ``content`` is the raw bytes - ``_response_to_dict`` then falls back to
+    ``json.loads``.
+
+    TODO(ionq-core): transitional. Drop this helper (and ``_BareResponse``
```
guenp, line 114: Should we track these in a github issue (and link GH issue # in the todo?)

qiskit_ionq/ionq_client.py
```diff
@@ -24,28 +24,144 @@
...
+        _IonQCoreAPIError(status_code=status, body=body_text, message=body_text)
+    )
+    if raise_retriable and exceptions._is_retriable("GET", status):
+        raise IonQRetriableError(err)
+    raise err
+
+
 class IonQClient:
-    """IonQ API Client
+    """IonQ API Client backed by ``ionq-core``.
```
guenp, line 164: I would leave it as-is

qiskit_ionq/ionq_client.py
```diff
@@ -66,13 +182,26 @@ def __init__(
             url = url[:-1]
         self._url = url
         self._user_agent = get_user_agent()
+        # ``ionq-core`` is the sole HTTP transport. ``api_key`` is required at
+        # construction time; the legacy ``IonQClient()`` with no token also
+        # constructed successfully and only failed on the first wire call, so
+        # use a placeholder here to preserve that pattern (real calls without
+        # a real key will surface a 401 from the server, same as before).
+        self._core = _IonQCoreClient(
+            api_key=token or os.environ.get("IONQ_API_KEY") or "MISSING_API_KEY",
```
guenp, line 191: Maybe create a `_get_api_key` helper function for this

qiskit_ionq/ionq_client.py
```diff
@@ -138,15 +265,21 @@ def submit_job(self, job: IonQJob) -> dict:
...
-        req_path = self.make_path("jobs")
-        res = requests.post(
-            req_path,
-            data=as_json,
-            headers=self.api_headers,
-            timeout=30,
+        body_dict = json.loads(as_json)
+        body_cls = (
+            JSONMultiCircuitJob
+            if body_dict.get("type") == "ionq.multi-circuit.v1"
```
guenp, line 271: Can this be a variable e.g. `JOB_TYPE_V1`?

qiskit_ionq/ionq_client.py
```diff
@@ -24,28 +24,144 @@
...
+    raw = getattr(response, "content", None)
+    if raw:
+        try:
+            return json.loads(raw)
+        except (ValueError, TypeError):
+            return {}
+    return {}
+
+
+def _call_through_core(endpoint_call, /, **kwargs):
```
guenp, line 101: I would name this differently, it's not immediately clear to be what this function does based on this naming

qiskit_ionq/ionq_client.py
```diff
@@ -24,28 +24,144 @@
...
+    """Run an ``ionq-core`` ``sync_detailed`` and tolerate strict-parse misses.
+
+    ``ionq-core`` 0.1.x's OpenAPI-generated ``from_dict`` models pop required
+    keys eagerly, so any test fixture that builds a partial mock job/results
```
guenp, line 105: I think we should fix the mock behavior rather than changing the actual client code to catch errors thrown because the mock doesn't represent the real payload. This way, you create a new code path for thr tests and the actual production code path doesn't get tested.

guenp (changes requested):
Took a quick first look, I would implement this a bit differently. Right now, it seems an awkward rewrite with lots of local imports at it doesn't feel very seamless. Do we perhaps need a core wrapper that makes integration easier? Or is there just naturally a lot of overlap between naming and abstractions?

Issue #250, opened 2026-05-07
AttributeError: SparseObservable missing .to_list() in qiskit_circ_to_ionq_circ


AviRubin:
### Information
- **Qiskit IonQ Provider version**: 1.0.2
- **API URL or version**: https://api.ionq.co/v0.3
- **Python version**: Python 3.12.13
- **Operating system**: macOS ProductVersion 14.3.1 BuildVersion 23G93

### What is the current behavior?
Using `SparsePauliOp` or a `PauliEvolutinGate` (and likely other operations involving `SparseObservable`) results in an `AttributeError`. The translation layer in `qiskit-ionq` attempts to call `.to_list()` on the operator object, but this method does not exist for the `SparseObservable` class. The specific error thrown is `AttributeError: 'qiskit.quantum_info.SparseObservable' object has no attribute 'to_list'` which traces back to `File "/qiskit-ionq/qiskit_ionq/helpers.py", line 288, in qiskit_circ_to_ionq_circ
terms = [term[0] for term in instruction.operator.to_list()]`

### Steps to reproduce the problem
1. Install qiskit and qiskit-ionq
2. Define a quantum circuit that includes a `SparsePauliOp` such as what is found [here](https://github.com/AviRubin/Quantum/blob/main/QPE_electrolyte_LiSulfone.py)
3. Initialize IonQ Cloud simulator `backend = provider.get_backend("simulator")`
4. Attempt to run the circuit
The execution fails with the above `AttributeError`

### What is the expected behavior?
The `qiskit-ionq` provider should seamlessly handle `SparseObservable` objects introduced in recent Qiskit versions.

### Suggested solutions
Implement a check in `qiskit_ionq/helpers.py`. If the instruction operator lacks a `.to_list()` method, use `SparsePauliOp.from_sparse_observable(operator)` to cast it to a compatible type.

Based on release notes I think this issue either started in qiskit version 1.2 or 1.3, but maybe `SparseObservable` wasn't made the default `operator` until later.

AviRubin:
[Fixed](https://github.com/qiskit-community/qiskit-ionq/pull/251)

PR #251, opened 2026-05-07
Fix SparseObservable compatibility

AviRubin:
### Summary
Resolves an `AttributeError` caused by Qiskit versions introducing `SparseObservable` as the default `operator` in the `PauliEvolutionGate`, which lacks the `.to_list()` method. Implemented a check and used `SparsePauliOp.from_sparse_observable()` to cast the operator for backwards compatibility.

### Details and comments
More details in [Issue 250](https://github.com/qiskit-community/qiskit-ionq/issues/250)

### DIFF
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -275,12 +275,17 @@ def qiskit_circ_to_ionq_circ(
             )
 
         if instruction_name == "pauliexp":
+            operator = instruction.operator
+            # Patch for Qiskit 1.2+: Convert SparseObservable to SparsePauliOp
+            if not hasattr(operator, "to_list"):
+                from qiskit.quantum_info import SparsePauliOp
+                operator = SparsePauliOp.from_sparse_observable(operator)
             imag_coeff = any(coeff.imag for coeff in instruction.operator.coeffs)
             assert not imag_coeff, (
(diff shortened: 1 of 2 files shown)


splch:
Thank you for the PR! This is a great catch. I have a couple of requests before approval though

splch:
Two things:

1. **Coeff/term mismatch.** Lines 280 and 313 still read `instruction.operator.coeffs`, but `terms` on line 287 reads from the cast local `operator`. For `SparseObservable`s with projector/ladder terms, `from_sparse_observable` expands the term count and re-weights coeffs, so the API payload ends up with mismatched `terms` and `coefficients` lengths. Please switch both `coeffs` reads to the local `operator`.

2. **Regression test.** Could you add a sibling to `test_pauliexp_circuit` in `test/helpers/test_gate_serialization.py` that builds `PauliEvolutionGate` directly from a `SparseObservable`? The existing test only covers the `SparsePauliOp` path, so it wouldn't have caught this and won't catch its return.

Thanks!

AviRubin:
Nice catch on that expansion mismatch.

1. Switched the `coeffs` reads to use the local `operator` to ensure parity.
2. Added `test_pauliexp_circuit_sparse_observable` to `test_gate_serialization.py` to cover this path.

Thanks!

splch (approved):
LGTM!

PR #252, opened 2026-05-11
Simplifing the conftest.py a bit

1ucian0:
I was running the tests and I noticed that the `test/conftest.py` file could be slightly simpler.


splch:
Thank you @1ucian0!

splch (approved):

PR #257, opened 2026-06-01
fix: IonQBackend.status() across simulator, meta-QPU, and real QPUs

splch:
Three pre-existing bugs (all in v1.0.2) along the `IonQBackend.status()` path, surfaced during live v1.0.3 verification. Each has its own regression test.

### Bug 1 — simulator reports `status() == False`
`calibration()` returns `None` for simulators by design, and `bool(None and ...)` short-circuits to False. Fix: shortcut on `self._simulator`.

### Bug 2 — `get_calibration_data(limit=1)` crashes on empty backends
```
GET /v0.4/backends/qpu.qpu/characterizations -> {"characterizations": null}
```
[OpenAPI](https://github.com/ionq/ionq-core/blob/main/openapi.json) declares the field non-nullable, but the live API ships `null` for backends with no data. `dict.get(key, [])` returns `None` (not `[]`) when the key is present with a null value, so `chars[0]` raises `TypeError`. Fix: `chars or []`, return `None` when `limit=1` and empty. Return type widened to `... | None`.

### Bug 3 — every real QPU's `status()` crashes with `KeyError: 'status'`
`Characterization` has no `status` field per the spec (required keys: `{backend, connectivity, date, fidelity, id, qubits, timing}`), but `Characterization.status` did `self._data["status"]`. `getattr` only catches `AttributeError`, not `KeyError`. Fix: `.get("status", "available")`.

## Live verification

| backend | before | after |
|---|---|---|
| `ionq_simulator` | `False` | `True` |
| `ionq_qpu` (meta) | `TypeError` | `False` |
| `ionq_qpu.aria-1` | `KeyError` | `True` |
| `ionq_qpu.forte-1` | `KeyError` | `True` |

## Tests
- `test_simulator_status_is_true` (Bug 1) — replaces the buggy `test_status_dummy_response` that asserted `is False`
- `test_get_cal_data_no_chars` (Bug 2)
- `test_qpu_status_with_real_cal` (Bug 3) — mocks a spec-shaped `Characterization` with no `status` field

Each verified to fail without its production fix. **395 passed**; ruff/format/mypy/pylint clean.

## Sweep for similar bugs
`Characterization.connectivity`/`fidelity`/`timing` use the same `.get(key, default)` pattern as Bug 2 and would break if the API ever shipped those as null. Spec says required and the live aria-1 response has them populated, so not fixed speculatively. No other unguarded dict-property + `getattr` combinations found.

### DIFF
--- a/qiskit_ionq/ionq_backend.py
+++ b/qiskit_ionq/ionq_backend.py
@@ -265,7 +265,15 @@ class IonQBackend(Backend):
         return self.client.get_calibration_data(self._api_backend_name, limit=1)
 
     def status(self) -> bool:
-        """True if the backend is currently available."""
+        """True if the backend is currently available.
+
+        For simulators we short-circuit to ``True`` because ``calibration()``
+        returns ``None`` by design (the simulator has no characterization
+        endpoint), and the simulator is always available as long as the API
+        itself is reachable.
+        """
+        if self._simulator:
+            return True
         cal = self.calibration()
         return bool(cal and getattr(cal, "status", "available") == "available")
 
--- a/qiskit_ionq/ionq_backend.py
+++ b/qiskit_ionq/ionq_backend.py
@@ -267,10 +267,8 @@ class IonQBackend(Backend):
     def status(self) -> bool:
         """True if the backend is currently available.
 
-        For simulators we short-circuit to ``True`` because ``calibration()``
-        returns ``None`` by design (the simulator has no characterization
-        endpoint), and the simulator is always available as long as the API
-        itself is reachable.
+        Simulators short-circuit to ``True``: they have no characterization
+        endpoint but are reachable whenever the API is.
         """
         if self._simulator:
             return True
--- a/qiskit_ionq/ionq_backend.py
+++ b/qiskit_ionq/ionq_backend.py
@@ -262,7 +262,7 @@ class IonQBackend(Backend):
         """Return the latest characterization data (None for simulator)."""
         if self._simulator:
             return None
-        return self.client.get_calibration_data(self._api_backend_name, limit=1)
+        return self.client.get_latest_calibration(self._api_backend_name)
 
     def status(self) -> bool:
         """True if the backend is currently available.
--- a/qiskit_ionq/ionq_client.py
+++ b/qiskit_ionq/ionq_client.py
@@ -233,20 +233,15 @@ class IonQClient:
             IonQRetriableError: When a retriable error occurs during the request.
 
         Returns:
-            Characterization: A single instance when ``limit == 1`` and data
+            A single ``Characterization`` when ``limit == 1`` and data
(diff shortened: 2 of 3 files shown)


test/ionq_backend/test_base_backend.py
```diff
@@ -39,15 +39,54 @@
 from .. import conftest
 
 
-def test_status_dummy_response(mock_backend):
-    """Test that the IonQBackend returns an aribtrary backend status.
+def test_simulator_status_is_true(mock_backend):
+    """``status()`` must not conflate "has calibration data" with "is
+    available". MockBackend is ``simulator=True``.
+    """
+    assert mock_backend.status() is True
```
natestemen, line 44: Is there a reason you don't have to mock the request here? I ask partially because the request mocking in the two tests below help me understand the context, but this one seems to assume some specific state when a simulator is used. That might be fine, however.

qiskit_ionq/ionq_backend.py
```diff
@@ -265,7 +265,13 @@ def calibration(self) -> Characterization | None:
         return self.client.get_calibration_data(self._api_backend_name, limit=1)
 
     def status(self) -> bool:
-        """True if the backend is currently available."""
+        """True if the backend is currently available.
+
+        Simulators short-circuit to ``True``: they have no characterization
+        endpoint but are reachable whenever the API is.
```
natestemen, line 271: Question/out of scope for this PR, but: What is contained in the characterization data? If it's QPU characterization data, having "ideal" characterization for simulators seems handy for users who are trying to understand how to work with the data.

qiskit_ionq/ionq_client.py
natestemen: This function's interface is a bit awkward IMO.

1. `limit` is an `int`, but acts as a boolean flag. I would think if I pass `limit=2`, then I get two results, but that is not the case.
2. Would it make sense to always return `list[Characterization]`? Callers can index if they want a single item, and an empty list handles the no-data case more cleanly than None.

If we want an easy way to fetch the lastest calibration, maybe we add a `get_latest_calibration` function. WDYT?

natestemen (approved):
I have some questions/comments that are maybe a bit bigger than this PR, but no blockers on merging without addressing in full.

qiskit_ionq/ionq_client.py
splch: great catch! yes i agree and the API returns a list regardless so we shouldn't do anything fancy / one-off here

test/ionq_backend/test_base_backend.py
splch: good question - status() short-circuits on self._simulator before any HTTP, so no mock is needed. i expanded the docstring to make that clear and avoided asserting call_count == 0 since it'd break for the wrong reason if simulators ever gain characterizations

PR #258, opened 2026-06-01
feat: mid-circuit measurement via ionq.qasm3.v1

splch:
## What

Adds **mid-circuit measurement** and adopts the IonQ **artifacts endpoint** for results and compiled circuits the legacy per-`lang` endpoints no longer serve.

```python
qr, mid, result = QuantumRegister(1, "q"), ClassicalRegister(1, "mid"), ClassicalRegister(2, "result")
qc = QuantumCircuit(qr, mid, result)
qc.h(0)
qc.measure(0, mid[0])      # mid-circuit measurement
qc.x(0)
qc.measure(0, result[0])   # qubit reused after measurement
qc.x(0)
qc.measure(0, result[1])

res = backend.run(qc, shots=100, memory=True).result()
res.get_counts()   # register-split, e.g. {'00 0': 75, '01 1': 25}
```

## How

- **Routing** — `circuit_requires_qasm3()` detects qubit-reuse-after-measure, `reset`, and control flow; such circuits serialize to `ionq.qasm3.v1` (OpenQASM 3 via `qiskit.qasm3.dumps`) instead of raising. Everything else is unchanged (`ionq.circuit.v1`).
- **Results** — per-register shots from the v2 artifact are folded into Qiskit register-split `get_counts()`/`get_memory()` via the header's `clbit_labels`; the system-appended `output_all` is excluded.
- **Compiled circuits** — `/jobs/{id}/circuits/{lang}` now returns **410 Gone** (prod + staging); `compiled_circuit(lang)` resolves the format under `output.compilation.compiled_circuits` and fetches it via the shared `get_artifact()` primitive (the dead client method is removed).

## Notes / guards

- Single-circuit only; QASM 3 executes on the simulator and MCM-capable QPUs (Tempo) and is rejected elsewhere server-side.
- Classical-register names that `qasm3.dumps` renames to dodge OpenQASM 3 reserved words (`output` → `output_0`) are rejected at submission — a silent rename would corrupt that register's results.
- `_build_settings()` assembles the settings/`error_mitigation` block for both submission paths (also fixes a v1/qasm3 divergence on `symmetry_verification`).

## Validation

413 tests pass (incl. live end-to-end against the IonQ API where the job type is enabled); pylint 10.00/10; ruff + ruff-format + mypy clean. Prod has 410'd the old circuits endpoint but its compiler doesn't emit the artifacts yet (`compiled_circuits: null`), so `compiled_circuit()` raises a clear "no compiled circuit" error there until the API side finishes rolling out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ The ideal simulator does not produce per-shot data; calling `get_memory()` on a 
 
 ### Mid-circuit measurement
 
-Circuits that reuse a qubit after measuring it — plus mid-circuit `reset` and classical control flow — are submitted automatically as OpenQASM 3 (`ionq.qasm3.v1`); no extra flags needed. Outcomes are reported per declared classical register, matching Qiskit's usual register-split formatting. Single-circuit submissions only; pass `error_mitigation`/`symmetry_verification`, `include_leakage`, or `verbatim` via `job_settings`.
+Circuits that reuse a qubit after measuring it — plus mid-circuit `reset` and classical control flow — are submitted automatically as OpenQASM 3 (`ionq.qasm3.v1`); no extra flags needed. Outcomes are reported per declared classical register, matching Qiskit's usual register-split formatting. Single-circuit submissions only; pass `error_mitigation`/`symmetry_verification`, `include_leakage`, or `verbatim` via `job_settings`. OpenQASM 3 jobs run on the simulator and mid-circuit-measurement-capable QPUs (Tempo); other targets reject them server-side. Classical-register names that collide with OpenQASM 3 reserved words (e.g. `output`, `input`, `measure`) are rejected at submission — rename them.
 
 ```python
 from qiskit import QuantumCircuit, QuantumRegister, ClassicalRegister
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ job = backend.run(qc, dry_run=True)
 job.wait_for_final_state()
 
 native = job.compiled_circuit(lang="native")   # IonQ-native gate JSON (dict)
-qasm3  = job.compiled_circuit(lang="qasm3")    # OpenQASM 3 source (str)
+ore    = job.compiled_circuit(lang="ore")      # lower-level ORE JSON (dict)
 ```
 
 The compiled circuit is fetched from the job's published artifacts (`output.compilation.compiled_circuits`); `lang` is matched against the available format keys (e.g. `"native"` → `ionq.native.v1`). The available formats vary by job and compiler — if the requested one isn't published, `IonQJobError` lists what is. Dry-run jobs produce no measurement results, so calling `job.result()` on one raises `IonQJobError` directing you to `compiled_circuit(...)`.
@@ -117,9 +117,11 @@ memory = job.get_memory()       # ['11', '00', '11', '00', ...]
 
 The ideal simulator does not produce per-shot data; calling `get_memory()` on a job submitted without `memory=True` raises `IonQBackendError`.
 
-### Mid-circuit measurement
+### Mid-circuit measurements
 
-Circuits that reuse a qubit after measuring it — plus mid-circuit `reset` and classical control flow — are submitted automatically as OpenQASM 3 (`ionq.qasm3.v1`); no extra flags needed. Outcomes are reported per declared classical register, matching Qiskit's usual register-split formatting. Single-circuit submissions only; pass `error_mitigation`/`symmetry_verification`, `include_leakage`, or `verbatim` via `job_settings`. OpenQASM 3 jobs run on the simulator and mid-circuit-measurement-capable QPUs (Tempo); other targets reject them server-side. Classical-register names that collide with OpenQASM 3 reserved words (e.g. `output`, `input`, `measure`) are rejected at submission — rename them.
+The IonQ provider supports circuits with mid-circuit measurements, qubit reuse after measurement, mid-circuit `reset`, and classical control flow. Outcomes are reported per declared classical register, matching Qiskit's usual register-split formatting. Single-circuit submissions only; pass `error_mitigation`/`symmetry_verification`, `include_leakage`, or `verbatim` via `job_settings`.
+
+Such circuits are submitted automatically as OpenQASM 3 (`ionq.qasm3.v1`) — no extra flags needed. OpenQASM 3 jobs run on the simulator and mid-circuit-measurement-capable QPUs (Tempo); other targets reject them server-side. Classical-register names that collide with OpenQASM 3 reserved words (e.g. `output`, `input`, `measure`) are rejected at submission — rename them.
 
 ```python
 from qiskit import QuantumCircuit, QuantumRegister, ClassicalRegister
@@ -130,16 +132,17 @@ result = ClassicalRegister(2, "result")
 qc = QuantumCircuit(qr, mid, result)
 
 qc.h(0)
-qc.measure(0, mid[0])      # mid-circuit measurement
-qc.x(0)
-qc.measure(0, result[0])   # qubit reused after measurement
+qc.measure(0, mid[0])          # mid-circuit measurement
+with qc.if_test((mid, 1)):      # classical control flow on the outcome
+    qc.x(0)
+qc.measure(0, result[0])       # qubit reused after measurement
 qc.x(0)
 qc.measure(0, result[1])
 
 job = backend.run(qc, shots=100, memory=True)
 result = job.result()
-result.get_counts()        # split across registers, e.g. {'00 0': 75, '01 1': 25}
-result.get_memory()        # per-shot, e.g. ['00 0', '01 1', ...]
+result.get_counts()        # split across registers, e.g. {'10 0': 50, '10 1': 50}
+result.get_memory()        # per-shot, e.g. ['10 0', '10 1', ...]
 ```
 
 ### Basis gates and transpilation
--- a/README.md
+++ b/README.md
@@ -100,11 +100,11 @@ backend = provider.get_backend("ionq_qpu.forte-1")
 job = backend.run(qc, dry_run=True)
 job.wait_for_final_state()
 
-native_json = job.compiled_circuit(lang="native")   # IonQ-native JSON
-qasm3       = job.compiled_circuit(lang="qasm3")    # OpenQASM 3
+native = job.compiled_circuit(lang="native")   # IonQ-native gate JSON (dict)
+qasm3  = job.compiled_circuit(lang="qasm3")    # OpenQASM 3 source (str)
 ```
 
-Dry-run jobs produce no measurement results, so calling `job.result()` on one raises `IonQJobError` directing you to `compiled_circuit(...)`.
+The compiled circuit is fetched from the job's published artifacts (`output.compilation.compiled_circuits`); `lang` is matched against the available format keys (e.g. `"native"` → `ionq.native.v1`). The available formats vary by job and compiler — if the requested one isn't published, `IonQJobError` lists what is. Dry-run jobs produce no measurement results, so calling `job.result()` on one raises `IonQJobError` directing you to `compiled_circuit(...)`.
 
 ### Per-shot memory (`memory`)
 
--- a/README.md
+++ b/README.md
@@ -117,6 +117,31 @@ memory = job.get_memory()       # ['11', '00', '11', '00', ...]
 
 The ideal simulator does not produce per-shot data; calling `get_memory()` on a job submitted without `memory=True` raises `IonQBackendError`.
 
+### Mid-circuit measurement
+
+Circuits that reuse a qubit after measuring it — plus mid-circuit `reset` and classical control flow — are submitted automatically as OpenQASM 3 (`ionq.qasm3.v1`); no extra flags needed. Outcomes are reported per declared classical register, matching Qiskit's usual register-split formatting. Single-circuit submissions only; pass `error_mitigation`/`symmetry_verification`, `include_leakage`, or `verbatim` via `job_settings`.
+
+```python
+from qiskit import QuantumCircuit, QuantumRegister, ClassicalRegister
+
+qr = QuantumRegister(1, "q")
+mid = ClassicalRegister(1, "mid")
+result = ClassicalRegister(2, "result")
+qc = QuantumCircuit(qr, mid, result)
+
+qc.h(0)
+qc.measure(0, mid[0])      # mid-circuit measurement
+qc.x(0)
+qc.measure(0, result[0])   # qubit reused after measurement
+qc.x(0)
+qc.measure(0, result[1])
+
+job = backend.run(qc, shots=100, memory=True)
+result = job.result()
+result.get_counts()        # split across registers, e.g. {'00 0': 75, '01 1': 25}
+result.get_memory()        # per-shot, e.g. ['00 0', '01 1', ...]
+```
+
 ### Basis gates and transpilation
 
 The IonQ provider provides access to the full IonQ Cloud backend, which includes its own transpilation and compilation pipeline. As such, IonQ provider backends have a broad set of "basis gates" that they will accept — effectively anything the IonQ API will accept. The current supported gates can be found [on our docs site](https://docs.ionq.com/#tag/quantum_programs).
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -50,6 +50,7 @@ from qiskit import __version__ as qiskit_version
 from qiskit.user_config import get_config
 from qiskit.circuit import (
     controlledgate as q_cgates,
+    ControlFlowOp,
     QuantumCircuit,
     QuantumRegister,
     ClassicalRegister,
@@ -162,13 +163,13 @@ def circuit_requires_qasm3(input_circuit: QuantumCircuit) -> bool:
     """True for mid-circuit measurement, reset, or control flow (not
     expressible as a flat v1 gate list, so submitted as OpenQASM 3).
     """
-    control_flow = {"if_else", "while_loop", "for_loop", "switch_case"}
     measured: set[int] = set()
     for inst in input_circuit.data:
-        name = inst.operation.name
+        operation = inst.operation
+        name = operation.name
         if name in ("barrier", "delay"):
             continue
-        if name in control_flow:
+        if isinstance(operation, ControlFlowOp):
             return True
         if name == "reset":
             return True
@@ -476,7 +477,7 @@ def _qasm3_data(circuit: QuantumCircuit) -> str:
     if renamed:
         raise ionq_exceptions.IonQJobError(
             f"Classical register name(s) {renamed} are OpenQASM 3 reserved "
-            "words and get renamed on export; rename them and resubmit."
+            "words; rename them and resubmit."
         )
     return data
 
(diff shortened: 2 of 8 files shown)


Cynocracy (approved):
Main question is what happens if you try to submit QASM to a system that doesn't support it as an input, or are we able to compat-layer that?

Otherwise LGTM!

duvanmoionq:
> Main question is what happens if you try to submit QASM to a system that doesn't support it as an input, or are we able to compat-layer that?
> 
> Otherwise LGTM!

@Cynocracy I merged a new change already that restrict qasm3 to only redwood || simulator otherwise it will failed the job.  that should be enough for your concern right?

Screen Shot 2026-06-03 at 5 31 02 PM

Cynocracy:
@duvanmoionq yup that answers my question satisfactory! thank you :)

README.md
```diff
@@ -117,6 +117,31 @@ memory = job.get_memory()       # ['11', '00', '11', '00', ...]
 
 The ideal simulator does not produce per-shot data; calling `get_memory()` on a job submitted without `memory=True` raises `IonQBackendError`.
 
+### Mid-circuit measurement
```
antalszava: (very minor)
```suggestion
### Mid-circuit measurements
```

test/ionq_job/test_job.py
```diff
@@ -1141,6 +1141,27 @@ def _dry_run_job_response(job_id, target="qpu.forte-1"):
...
+                "precision": "1E-3",
+                "gate_basis": "ZZ",
+                "service_version": "v0.4",
+                "compiled_circuits": {
+                    "ionq.native.v1": {
+                        "id": "native-aid",
+                        "format": "ionq.native.v1",
+                        "media_type": "application/json",
+                    },
+                    "ionq.qasm3.v1": {
```
duvanmoionq: So currently we are not generating `ionq.qasm3.v1` compilation,  for a `qasm3` we only generate 
```
"ionq.mir.v1": {
                    "id": "bd76df27-a4ab-49ca-8888-e3d9115bd1b0",
                    "format": "ionq.mir.v1",
                    "media_type": "application/octet-stream"
                }
```

duvanmoionq: for a `qis` we generate this 
```
"ionq.native.v1": {
                    "id": "f6fab9a8-2f9a-4035-9250-bf5161007de9",
                    "format": "ionq.native.v1",
                    "media_type": "application/json"
                },
                "ionq.ore.v1": {
                    "id": "0d6857ce-f2d6-4b30-9a6d-433633e92854",
                    "format": "ionq.ore.v1",
                    "media_type": "application/json"
                }
```

README.md
```diff
@@ -117,6 +117,31 @@ memory = job.get_memory()       # ['11', '00', '11', '00', ...]
...
+
+Circuits that reuse a qubit after measuring it — plus mid-circuit `reset` and classical control flow — are submitted automatically as OpenQASM 3 (`ionq.qasm3.v1`); no extra flags needed. Outcomes are reported per declared classical register, matching Qiskit's usual register-split formatting. Single-circuit submissions only; pass `error_mitigation`/`symmetry_verification`, `include_leakage`, or `verbatim` via `job_settings`. OpenQASM 3 jobs run on the simulator and mid-circuit-measurement-capable QPUs (Tempo); other targets reject them server-side. Classical-register names that collide with OpenQASM 3 reserved words (e.g. `output`, `input`, `measure`) are rejected at submission — rename them.
```
antalszava: imho it's good to rephrase this start of the paragraph a bit:
* First: mention that mid-circuit measurements are supported;
* Second: personally, I'd distinguish mcm, qubit reuse and the likes, mention them separately and then mention the point about using OpenQASM 3, as they are different features. Imo this is important because e.g., "simple mcm" can work without qubit reuse, control flow, etc..
* Push down OpenQASM 3 note: users don't **need** to know about this detail, until it becomes important because of the target constraints (simulator + Tempo).

```suggestion
The IonQ provider supports circuits with mid-circuit measurements, qubit reuse and further instructions like a mid-circuit `reset` or classical control flow. Outcomes are reported per declared classical register, matching Qiskit's usual register-split formatting. Single-circuit submissions only; pass `error_mitigation`/`symmetry_verification`, `include_leakage`, or `verbatim` via `job_settings`.

Note that circuits that contain mid-circuit measurements are submitted automatically as OpenQASM 3 (`ionq.qasm3.v1`) - no extra flags needed. OpenQASM 3 jobs run on the simulator and mid-circuit-measurement-capable QPUs (Tempo); other targets reject them server-side. Classical-register names that collide with OpenQASM 3 reserved words (e.g. `output`, `input`, `measure`) are rejected at submission — rename them.
```

README.md
```diff
@@ -117,6 +117,31 @@ memory = job.get_memory()       # ['11', '00', '11', '00', ...]
...
+mid = ClassicalRegister(1, "mid")
+result = ClassicalRegister(2, "result")
+qc = QuantumCircuit(qr, mid, result)
+
+qc.h(0)
+qc.measure(0, mid[0])      # mid-circuit measurement
+qc.x(0)
+qc.measure(0, result[0])   # qubit reused after measurement
+qc.x(0)
+qc.measure(0, result[1])
```
antalszava, line 140: imho good to add an example for control flow too, if fitting and possibly keeping the example still succinct

qiskit_ionq/ionq_job.py
```diff
@@ -250,39 +253,49 @@ def dry_run(self) -> bool:
...
             raise exceptions.IonQJobStateError(
                 "Cannot fetch compiled circuit until the job reaches a final state. "
                 "Call wait_for_final_state() first."
             )
         assert self._job_id is not None
-        return self._client.get_compiled_circuit(self._job_id, lang=lang)
+        circuits = ((self._metadata.get("output") or {}).get("compilation") or {}).get(
+            "compiled_circuits"
+        ) or {}
+        fmt = self._resolve_compiled_format(lang, circuits)
```
duvanmoionq, line 292: @splch here `circuits` will be something like 
```
{
                "ionq.mir.v1": {
                    "id": "13ed34d1-1097-4daa-9524-23fadac3124b",
                    "format": "ionq.mir.v1",
                    "media_type": "application/octet-stream"
                }
            }
```
it is an object 
so so you can do like 
```
fmt = circuits[lang]
```
assuming `lang` =  `ionq.mir.v1` that should be enough but no sure if I am missing anything!

qiskit_ionq/helpers.py
```diff
@@ -157,6 +158,29 @@ def native_2q_gate(value: str | None) -> Literal["ms", "zz"] | None:
     return None
 
 
+def circuit_requires_qasm3(input_circuit: QuantumCircuit) -> bool:
+    """True for mid-circuit measurement, reset, or control flow (not
+    expressible as a flat v1 gate list, so submitted as OpenQASM 3).
+    """
+    control_flow = {"if_else", "while_loop", "for_loop", "switch_case"}
```
antalszava: could `qiskit.circuit.ControlFlowOp` be used here?

qiskit_ionq/helpers.py
```diff
@@ -157,6 +158,29 @@ def native_2q_gate(value: str | None) -> Literal["ms", "zz"] | None:
     return None
 
 
+def circuit_requires_qasm3(input_circuit: QuantumCircuit) -> bool:
```
antalszava, line 162: could this be used instead (perhaps not which is fine)?

```py

from qiskit.qasm2 import QasmError

def circuit_requires_qasm3(circuit):
    try:
        # Attempt an OpenQASM 2 export
        circuit.qasm()
        return False # It can be represented in QASM 2
    except QasmError:
        return True  # It has features that force QASM 3
```

qiskit_ionq/ionq_job.py
```diff
@@ -487,6 +508,8 @@ def status(self, detailed: bool = False) -> jobstatus.JobStatus | dict:
                     for k, v in results.items()
                     if isinstance(v, dict) and isinstance(v.get("url"), str)
                 }
+                if self._is_qasm3:
+                    self._shots_artifact_id = self._find_shots_v2_id(results)
```
duvanmoionq: Ideally here you can have a enum with all the formats we support. 
results will be something like this 
```
{
        "shots": {
            "url": "/v0.4/jobs/019e93b8-c769-7142-a8ca-5a8c8891c1b5/results/shots"
        },
        "histogram": {
            "url": "/v0.4/jobs/019e93b8-c769-7142-a8ca-5a8c8891c1b5/results/histogram"
        },
        "probabilities": {
            "url": "/v0.4/jobs/019e93b8-c769-7142-a8ca-5a8c8891c1b5/results/probabilities"
        },
        "ionq.result.shots.json.v2": {
            "id": "be1e3413-3a05-4c3b-b8f5-c47675bfd9ff",
            "format": "ionq.result.shots.json.v2",
            "media_type": "application/json"
        },
        "ionq.result.histogram.json.v2": {
            "id": "1e29cf34-5a27-498b-bf2a-f4c34344fee7",
            "format": "ionq.result.histogram.json.v2",
            "media_type": "application/json"
        },
        "ionq.result.probabilities.json.v2": {
            "id": "b3c62fdf-cff1-427f-8460-ec03d5bdeac0",
            "format": "ionq.result.probabilities.json.v2",
            "media_type": "application/json"
        }
```

```
class ResultFormat(StrEnum):
    """Result wire formats shared across execution and job artifacts."""

    PROBABILITIES_LEGACY = "ionq.result.probabilities.json.v1"
    SHOTS_LEGACY = "ionq.result.shots.json.v1"
    SHOTS_V2 = "ionq.result.shots.json.v2"
    HISTOGRAM_V2 = "ionq.result.histogram.json.v2"
    PROBABILITIES_V2 = "ionq.result.probabilities.json.v2"
```

so you can just do `results[ResultFormat.PROBABILITIES_V2]` that's it will be easier.

qiskit_ionq/ionq_job.py
```diff
@@ -641,6 +664,90 @@ def _raw_shots_per_circuit(self) -> list[list | None]:
             per_circuit.append(self._fetch_raw_shots(url, ctx=f"child {child_id}"))
         return per_circuit
 
+    @staticmethod
+    def _find_shots_v2_id(results: dict) -> str | None:
```
duvanmoionq: based on the comment above you don't need `_find_shots_v2_id`

qiskit_ionq/helpers.py
```diff
@@ -435,6 +459,108 @@ def decompress_metadata_string(
...
+        dumps as _qasm3_dumps,
+    )
+
+    data = _qasm3_dumps(circuit)
+    emitted = set(re.findall(r"\bbit(?:\[\d+\])?\s+(\w+)\s*;", data))
+    renamed = [creg.name for creg in circuit.cregs if creg.name not in emitted]
+    if renamed:
+        raise ionq_exceptions.IonQJobError(
+            f"Classical register name(s) {renamed} are OpenQASM 3 reserved "
+            "words and get renamed on export; rename them and resubmit."
```
antalszava: what does `and get renamed on export` mean? imho good to clarify or just drop it, msg is clear without it

qiskit_ionq/ionq_job.py
```diff
@@ -641,6 +664,90 @@ def _raw_shots_per_circuit(self) -> list[list | None]:
...
+        """Artifact id of the v2 per-register shots, or ``None``."""
+        for value in results.values():
+            if isinstance(value, dict) and str(value.get("format", "")).endswith(
+                "shots.json.v2"
+            ):
+                return value.get("id")
```
antalszava: this now iterates all results and gets the id of the very first v2 formatted value - could we have more than 1 such value? If so is this the right behaviour?

qiskit_ionq/ionq_job.py
```diff
@@ -641,6 +664,90 @@ def _raw_shots_per_circuit(self) -> list[list | None]:
...
+        return None
+
+    def _fetch_qasm3_shots(self, extra_query_params: dict | None = None) -> list:
+        """Fetch the per-register shots array for a qasm3 (MCM) job."""
+        if not self._shots_artifact_id:
+            raise exceptions.IonQJobError(
+                f"Job {self._job_id} reported no per-register shots artifact; "
+                "cannot build mid-circuit measurement results."
+            )
+        assert self._job_id is not None
```
antalszava, line 678: imho an error should preferred instead of an assert if this can feasibly happen to give a clear error message.

qiskit_ionq/ionq_job.py
```diff
@@ -411,12 +424,17 @@ def result(
                     "job.compiled_circuit(lang='native' or 'qasm3') to "
                     "retrieve the compiled circuit instead."
                 )
-            response = self._client.get_results(
-                results_url=self._results_urls.get("probabilities", ""),
-                sharpen=sharpen,
-                extra_query_params=extra_query_params,
-            )
-            self._result = self._format_result(response)
+            if self._is_qasm3:
```
antalszava, line 428: `self._is_qasm3` is set in `status`, but I'm not sure it is necessarily called up until now - is that right? can it happen that `self._is_qasm3` is not yet set when the execution hits here just because `status` has not yet been called?

qiskit_ionq/helpers.py
```diff
@@ -458,10 +584,31 @@ def qiskit_to_ionq(
...
+    multi_circuit = isinstance(circuit, (list, tuple))
+
+    # Route circuits the v1 gate list can't express (see circuit_requires_qasm3).
+    circuits_to_check = circuit if multi_circuit else [circuit]
+    if any(circuit_requires_qasm3(c) for c in circuits_to_check):
+        if multi_circuit:
+            raise ionq_exceptions.IonQJobError(
+                "Mid-circuit measurement, reset, and classical control flow are "
+                "only supported for single-circuit submissions. Submit each such "
+                "circuit as its own job."
```
antalszava: (minor for readability)
```suggestion
                "only supported for single-circuit submissions. Submit such circuits "
                "in separate jobs."
```

test/helpers/test_qiskit_to_ionq.py
```diff
@@ -463,3 +464,143 @@ def test_counts_marginalize(simulator_backend, requests_mock):
...
+    qc.measure(0, 0)
+    assert circuit_requires_qasm3(qc) is True
+
+
+def test_requires_qasm3_ctrl_flow():
+    """Classical control flow requires the qasm3 path."""
+    qc = QuantumCircuit(1, 1)
+    qc.h(0)
+    qc.measure(0, 0)
+    with qc.if_test((qc.cregs[0], 1)):
```
antalszava, line 503: would be good to have a test case for the else branch too

test/helpers/test_qiskit_to_ionq.py
```diff
@@ -463,3 +464,143 @@ def test_counts_marginalize(simulator_backend, requests_mock):
...
+    assert circuit_requires_qasm3(_mcm_circuit()) is True
+
+
+def test_requires_qasm3_reset():
+    """A mid-circuit reset requires the qasm3 path."""
+    qc = QuantumCircuit(1, 1)
+    qc.h(0)
+    qc.reset(0)
+    qc.measure(0, 0)
+    assert circuit_requires_qasm3(qc) is True
```
antalszava, line 495: perhaps good to test the case where we have a measurement followed by a gate (but no explicit terminal measurements)?

```py
from qiskit import QuantumCircuit

qc = QuantumCircuit(1, 1)
qc.h(0)
qc.measure(0, 0)  # Measurement here...
qc.x(0)           # ...followed by a gate. No final measurement.

print(qc)
```

test/helpers/test_qiskit_to_ionq.py
```diff
@@ -463,3 +464,143 @@ def test_counts_marginalize(simulator_backend, requests_mock):
...
+
+def test_qasm3_payload_shape(qpu_backend):
+    """MCM circuits serialize to an ionq.qasm3.v1 payload with QASM 3 input."""
+    ionq_json = qiskit_to_ionq(_mcm_circuit(), qpu_backend, passed_args={"shots": 10})
+    payload = json.loads(ionq_json)
+    assert payload["type"] == "ionq.qasm3.v1"
+    assert payload["input"]["qubits"] == 1
+    data = payload["input"]["data"]
+    assert data.startswith("OPENQASM 3.0;")
+    # Don't assert qiskit's exact qasm3 spelling; just the structure.
```
antalszava, line 546: could we still assert something like the number of expected gates to make sure that no unwanted gates/instructions are being added to the paylod?

antalszava (commented):
No blockers found after a skim through 💯

README.md
splch: done

README.md
splch: Reworked it — features up front, the OpenQASM 3 stuff in a second paragraph. Thanks for the wording.

README.md
splch: Added an `if_test` block to the example.

qiskit_ionq/helpers.py
splch: Tried that route but `.qasm()` is gone as of qiskit 1.0, and QASM 2 happily encodes measure/reset/if so it'd return False for the circuits we care about. Stuck with the structural check.

qiskit_ionq/helpers.py
splch: Good call, using `isinstance(op, ControlFlowOp)` now.

qiskit_ionq/helpers.py
splch: Dropped that bit, good point.

qiskit_ionq/helpers.py
splch: done

qiskit_ionq/ionq_job.py
splch: Yep, it's a direct key lookup now. Kept a thin short-name shim so `compiled_circuit("native")` etc. still works.

qiskit_ionq/ionq_job.py
splch: It's fine — `_is_qasm3` gets set in the same `status()` call that flips the job to DONE, and we only read it once we're DONE.

qiskit_ionq/ionq_job.py
splch: Done, added the enum and it's `results.get(ResultFormat.SHOTS_V2)` now.

qiskit_ionq/ionq_job.py
splch: Yep, gone.

qiskit_ionq/ionq_job.py
splch: Moot now — ripped out the scan for a single keyed lookup.

qiskit_ionq/ionq_job.py
splch: I think it's safe — `status()` never reaches a final state with no job id, so we can't get here without one. It also does the `str | None -> str` narrowing mypy needs. Can make it a raised error instead if you'd rather.

test/ionq_job/test_job.py
splch: Good catch, fixed. Fixture uses `ionq.native.v1`/`ionq.ore.v1` now, and I added a test for the qasm3 `ionq.mir.v1` case — since it's octet-stream the client hands it back as bytes instead of choking on json.

test/helpers/test_qiskit_to_ionq.py
splch: Added `test_requires_qasm3_reuse` for that.

test/helpers/test_qiskit_to_ionq.py
splch: Added one. Heads up that if/else is still a single `IfElseOp`, so same path as a plain if — can point it at a for/switch if you want a distinct op covered.

test/helpers/test_qiskit_to_ionq.py
splch: Added — checks 1 h / 2 x / 3 measure in the exported qasm so a stray gate would fail it.

splch:
Confirmed it surfaces cleanly on our end too — unsupported targets come back as `IonQJobFailureError` (`UnsupportedTarget`).

PR #260, opened 2026-06-05
fix: pin compilation.service_version v0.4 for OpenQASM 3 jobs (1.1.1)

splch:
## What

`1.1.1` patch. Fixes mid-circuit-measurement results coming back **all-zeros** on production, and removes mention of features the platform doesn't support yet.

## Root cause

MCM (`ionq.qasm3.v1`) jobs only return **register-named** per-register results when compiled by the **v0.4 compiler stack**. The SDK wasn't setting `settings.compilation.service_version`, so prod compiled with the older default and returned index-flattened registers (`c0`, `c1`, …) instead of `mid`/`result`. The name-based decoder then read every register as `0` → silently wrong counts. (The platform plans to make v0.4 the default "soon," but there's no date — so the client must pin it.)

## Fix

- **Pin `settings.compilation.service_version = "v0.4"` on the OpenQASM 3 path** (caller-overridable). Verified live on the prod simulator: registers come back as `mid`/`result`/`output_all` and `get_counts()` is correct (e.g. `{'01 0': 108, '10 1': 92}`).
- **Docs:** drop `control flow`, `include_leakage`, and `verbatim` (not supported in this phase); note MCM runs on the **simulator** today (per-register shots need a `noise_model`); QPU support is rolling out.
- Bump to `1.1.1`.

## Notes
- The decoder is unchanged and correct for v0.4 output; this is purely about asking the server for that output.
- Control-flow circuits are still routed to the OpenQASM 3 path (the only path that could ever express them); they're just no longer advertised.
- Tests cover the `service_version` default + caller override; full suite green (417), pylint/ruff/mypy clean.

### DIFF
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -160,8 +160,10 @@ def native_2q_gate(value: str | None) -> Literal["ms", "zz"] | None:
 
 
 def circuit_requires_qasm3(input_circuit: QuantumCircuit) -> bool:
-    """True for mid-circuit measurement, reset, or control flow (not
-    expressible as a flat v1 gate list, so submitted as OpenQASM 3).
+    """True for circuits the flat v1 gate list can't express -- mid-circuit
+    measurement, reset, or control flow -- which are submitted as OpenQASM 3.
+
+    Control flow is detected and routed, but isn't supported server-side yet.
     """
     measured: set[int] = set()
     for inst in input_circuit.data:
@@ -202,7 +204,9 @@ def qiskit_circ_to_ionq_circ(
 
     Raises:
         IonQGateError: If an unsupported instruction is supplied.
-        IonQMidCircuitMeasurementError: If a mid-circuit measurement is detected.
+        IonQMidCircuitMeasurementError: If a qubit is used after measurement. The
+          flat v1 format can't express this; ``backend.run`` routes such circuits
+          to OpenQASM 3, so this only fires when this builder is called directly.
         IonQPauliExponentialError: If non-commuting PauliExponentials are found without
           the appropriate flag.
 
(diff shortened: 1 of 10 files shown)


qiskit_ionq/version.py
```diff
@@ -34,7 +34,7 @@
 pkg_parent = pathlib.Path(__file__).parent.parent.absolute()
 
 # major, minor, patch
-VERSION_INFO = ".".join(map(str, (1, 1, 0)))
+VERSION_INFO = ".".join(map(str, (1, 1, 1)))
```
natestemen, line 37: What's up with this method of setting the version? (as opposed to "1.1.1")

qiskit_ionq/helpers.py
```diff
@@ -550,8 +550,12 @@ def _qiskit_to_qasm3_json(  # pylint: disable=too-many-positional-arguments
...
 
     settings = _build_settings(passed_args, backend)
-    if settings:
-        ionq_json["settings"] = settings
+    # OpenQASM 3 jobs need the v0.4 compiler stack to return register-named
+    # results; pin it unless the caller set their own service_version.
+    compilation = dict(settings.get("compilation") or {})
+    compilation.setdefault("service_version", "v0.4")
+    settings["compilation"] = compilation
+    ionq_json["settings"] = settings
```
natestemen, line 562: What was the previous default version?

natestemen (approved):
LGTM!

What do you think should be the longer term solution here? Should we continue setting the default version of the compiler, or is there a cleaner solution long term?

qiskit_ionq/version.py
splch: pre-existing style and i think it's easier when doing a dev release. i just do `, "dev0"`

qiskit_ionq/helpers.py
splch: v0.3 - no classical-register map (so registers came back c0/c1/… -> the all-zeros) and no compiled-circuit artifacts. v0.4 is the QASM3 / MCM stack

PR #261, opened 2026-06-09
Support Qiskit v1 and v2

basnijholt:
## Summary
- relax Qiskit dependency pins to support both v1 and v2
- add Qiskit compatibility helpers for result headers and MeasLevel
- add tox factors and CI matrix entries for Qiskit 1/2 across supported Python versions

## Tests
- tox -e py313-qiskit1,py313-qiskit2
- tox -e lint,docs
- git diff --check

### DIFF
--- a/qiskit_ionq/_qiskit_compat.py
+++ b/qiskit_ionq/_qiskit_compat.py
@@ -1,0 +1,82 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2017, 2018.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+# Copyright 2026 IonQ, Inc. (www.ionq.com)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Qiskit v1/v2 compatibility helpers."""
+
+from __future__ import annotations
+
+from typing import Any
+
+from qiskit.quantum_info import SparsePauliOp
+
+try:
+    from qiskit.result import MeasLevel
+except ImportError:
+    MEAS_LEVEL_CLASSIFIED = 2
+else:
+    MEAS_LEVEL_CLASSIFIED = MeasLevel.CLASSIFIED
+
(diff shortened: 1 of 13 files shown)


qiskit_ionq/helpers.py
```diff
@@ -325,7 +325,12 @@ def qiskit_circ_to_ionq_circ(
         if instruction_name == "pauliexp":
             operator = instruction.operator
             if not hasattr(operator, "to_list"):
-                operator = SparsePauliOp.from_sparse_observable(operator)
+                from_sparse_observable = getattr(
+                    SparsePauliOp, "from_sparse_observable", None
+                )
+                if from_sparse_observable is None:
+                    raise ionq_exceptions.IonQGateError(instruction_name, gateset)
```
natestemen: In qiskit v1, this will raise an error complaining that the operation is not supported on ionq backends:
https://github.com/qiskit-community/qiskit-ionq/blob/bd1284199ef24557ce53f036c4661e1218642cdf/qiskit_ionq/exceptions.py#L208-L210

but I think this is actually incorrect. The error should read something like "`PauliEvolutionGate` is not supported in qiskit v1. Please upgrade to qiskit v2 or use `SparsePauliOp`". This could send the user on a transpilation goose chase

qiskit_ionq/_qiskit_compat.py
```diff
@@ -0,0 +1,85 @@
...
+
+def normalize_result_headers(result: Any) -> Any:
+    """Normalize Qiskit Result experiment headers to dicts in v1 and v2."""
+    for experiment in getattr(result, "results", []):
+        header = getattr(experiment, "header", None)
+        if header is None or isinstance(header, dict):
+            continue
+        header_dict = header_to_dict(header)
+        if header_dict is None:
+            continue
```
natestemen: I think this is unreachable code. If `header` is None then you will hit the previous check and bail out.

qiskit_ionq/ionq_job.py
```diff
@@ -732,16 +733,18 @@ def _format_result_qasm3(self, shots: list):
...
-            {
-                "results": job_result,
-                "job_id": self.job_id(),
-                "backend_name": backend.name,
-                "backend_version": backend.backend_version,
-                "qobj_id": metadata.get("qobj_id"),
-                "success": success,
-                "time_taken": self._execution_time,
-            }
+        return normalize_result_headers(
```
natestemen: I want @splch to comment on this here, but in the case of adding a new function that returns results we have to remember to add `normalize_result_headers` which is easy to forget.

tox.ini
```diff
@@ -1,15 +1,17 @@
...
 [gh-actions]
 python =
-  3.10: py310, mypy
-  3.11: py311, mypy
-  3.12: py312, mypy
-  3.13: py313, mypy
+  3.10: py310-qiskit1, py310-qiskit2
+  3.11: py311-qiskit1, py311-qiskit2
+  3.12: py312-qiskit1, py312-qiskit2
+  3.13: py313-qiskit1, py313-qiskit2
```
natestemen: I'm not super familiar with tox, but claude is warning me that this is dead config now that the github action uses `tox-env`.

natestemen (commented):
I'm not comfortable making the final call here, but I did notice a few things during review.

I'll pass to @splch to the final call. Generally I'm supportive of broadening version requirements, especially for qiskit[^1], but if it adds significant complexity and we aren't getting any signals from users (i.e. github issues or relays from product team) they need this, then maybe it's overkill?

[^1]: I know the qiskit major versions have been a major pain for users. 20% of qiskit downloads are still `1.*` according to https://pepy.tech/projects/qiskit.

PR #263, opened 2026-06-24
feat: API-driven backend target (Tempo support, connectivity, capabilities)

splch:
Builds each backend's `Target` from the `/backends/{id}` API instead of hardcoded device data - groundwork for **Tempo** (`qpu.tempo-1` now resolves with no device-specific code).

- Replaced `get_n_qubits()` with `get_backend_config()` (one API call); deleted the `NATIVE_2Q_BY_FAMILY` map and `native_2q_gate()` heuristic.
- Qubit count, native 2q gate (`supported_native_gates`), and connectivity (→ coupling map) now come from the API.
- Added `supported_gates` / `supported_native_gates` / `supported_error_mitigations` properties.
- qasm3 (MCM) results: shots tagged with nonzero `leakage_bits` in the shots-v2 artifact are excluded from counts/memory, matching the platform's histogram/probabilities aggregation (verified against staging tempo-1, where the stub inflates every job by ~5% forced-leaked shots).

### DIFF
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -815,30 +815,26 @@ def api_backend_id(name: str) -> str:
     return f"qpu.{name}"
 
 
-def get_backend_config(
-    name: str, token: str | None = None, url: str | None = None
-) -> dict:
-    """Fetch a backend's static config from ``GET /backends/{id}``.
+def get_backends(token: str | None = None, url: str | None = None) -> dict[str, dict]:
+    """Fetch the ``GET /backends`` catalog, keyed by API id.
 
-    Includes qubit count, supported/native gates, and supported error-mitigation
-    techniques, so callers derive them from one request rather than hardcoding
-    per device.
-
-    Returns ``{}`` (and warns) when the endpoint is unreachable, so callers can
-    fall back to defaults.
+    One request returns every backend's static config (qubit count,
+    supported/native gates, supported error-mitigation techniques), so the
+    provider can resolve them all locally instead of querying per device.
+    Returns ``{}`` (and warns) when the endpoint is unreachable.
     """
     creds = resolve_credentials(token, url)
     headers = {"Authorization": f"apiKey {creds['token']}"} if creds["token"] else None
     try:
-        resp = requests.get(
-            f"{creds['url']}/backends/{api_backend_id(name)}",
-            headers=headers,
-            timeout=5,
-        )
+        resp = requests.get(f"{creds['url']}/backends", headers=headers, timeout=5)
         resp.raise_for_status()
-        return resp.json()
+        data = resp.json()
+        items = data if isinstance(data, list) else (data.get("backends") or [])
+        return {
+            b["backend"]: b for b in items if isinstance(b, dict) and "backend" in b
+        }
     except Exception as exception:  # pylint: disable=broad-except
-        warnings.warn(f"Failed to fetch config for {name}: {exception}.")
+        warnings.warn(f"Failed to fetch backends catalog: {exception}.")
         return {}
 
 
@@ -924,6 +920,7 @@ __all__ = [
     "decompress_metadata_string",
     "get_user_agent",
     "resolve_credentials",
-    "get_backend_config",
+    "api_backend_id",
+    "get_backends",
     "retry",
 ]
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -145,6 +145,7 @@ GATESET_MAP = {
 NATIVE_2Q_BY_FAMILY: dict[str, Literal["ms", "zz"]] = {
     "aria": "ms",
     "forte": "zz",
+    "tempo": "zz",
 }
 
 
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -854,21 +854,15 @@ def get_backends(token: str | None = None, url: str | None = None) -> dict[str, 
     One request returns every backend's static config (qubit count,
     supported/native gates, supported error-mitigation techniques), so the
     provider can resolve them all locally instead of querying per device.
-    Returns ``{}`` (and warns) when the endpoint is unreachable.
+    Failures raise; graceful fallbacks are the caller's responsibility.
     """
     creds = resolve_credentials(token, url)
     headers = {"Authorization": f"apiKey {creds['token']}"} if creds["token"] else None
-    try:
-        resp = requests.get(f"{creds['url']}/backends", headers=headers, timeout=5)
-        resp.raise_for_status()
-        data = resp.json()
-        items = data if isinstance(data, list) else (data.get("backends") or [])
-        return {
-            b["backend"]: b for b in items if isinstance(b, dict) and "backend" in b
-        }
-    except Exception as exception:  # pylint: disable=broad-except
-        warnings.warn(f"Failed to fetch backends catalog: {exception}.")
-        return {}
+    resp = requests.get(f"{creds['url']}/backends", headers=headers, timeout=5)
+    resp.raise_for_status()
+    data = resp.json()
+    items = data if isinstance(data, list) else (data.get("backends") or [])
+    return {b["backend"]: b for b in items if isinstance(b, dict) and "backend" in b}
 
 
 def retry(
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -803,11 +803,14 @@ def resolve_credentials(token: str | None = None, url: str | None = None) -> dic
     }
 
 
-def _api_backend_id(name: str) -> str:
-    """Local name -> API id (``ionq_qpu.forte-1``/``forte-1`` -> ``qpu.forte-1``,
-    ``ionq_simulator`` -> ``simulator``)."""
+def api_backend_id(name: str) -> str:
+    """Map a local backend name to its API id.
+
+    ``ionq_qpu.forte-1``/``forte-1`` -> ``qpu.forte-1``,
+    ``ionq_simulator`` -> ``simulator``, ``ionq_qpu`` -> ``qpu``.
+    """
     name = name.removeprefix("ionq_")
-    if name == "simulator" or name.startswith("qpu."):
+    if name in ("simulator", "qpu") or name.startswith("qpu."):
         return name
     return f"qpu.{name}"
 
@@ -815,15 +818,20 @@ def _api_backend_id(name: str) -> str:
 def get_backend_config(
     name: str, token: str | None = None, url: str | None = None
 ) -> dict:
-    """Fetch a backend's static config (qubits, supported/native gates, error
-    mitigations) from ``GET /backends/{id}``; ``{}`` (with a warning) if
-    unreachable so callers can fall back to defaults.
+    """Fetch a backend's static config from ``GET /backends/{id}``.
+
+    Includes qubit count, supported/native gates, and supported error-mitigation
+    techniques, so callers derive them from one request rather than hardcoding
+    per device.
+
+    Returns ``{}`` (and warns) when the endpoint is unreachable, so callers can
+    fall back to defaults.
     """
     creds = resolve_credentials(token, url)
     headers = {"Authorization": f"apiKey {creds['token']}"} if creds["token"] else None
     try:
         resp = requests.get(
-            f"{creds['url']}/backends/{_api_backend_id(name)}",
+            f"{creds['url']}/backends/{api_backend_id(name)}",
             headers=headers,
             timeout=5,
         )
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -457,6 +457,12 @@ def _qasm3_data(circuit: QuantumCircuit) -> str:
         dumps as _qasm3_dumps,
     )
 
+    # Drop the layout so dumps emits a virtual register (qubit[n] q;), which
+    # the API requires, not physical qubits ($0).
+    if getattr(circuit, "layout", None) is not None:
+        circuit = circuit.copy()
+        circuit._layout = None  # pylint: disable=protected-access
+
     data = _qasm3_dumps(circuit)
     emitted = set(re.findall(r"\bbit(?:\[\d+\])?\s+(\w+)\s*;", data))
     renamed = [creg.name for creg in circuit.cregs if creg.name not in emitted]
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -457,8 +457,9 @@ def _qasm3_data(circuit: QuantumCircuit) -> str:
         dumps as _qasm3_dumps,
     )
 
-    # Drop the layout so dumps emits a virtual register (qubit[n] q;), which
-    # the API requires, not physical qubits ($0).
+    # Drop the layout so dumps emits a virtual register (qubit[n] q;) instead
+    # of OpenQASM 3 physical-qubit references ($0, $1, ...), which the API
+    # rejects.
     if getattr(circuit, "layout", None) is not None:
         circuit = circuit.copy()
         circuit._layout = None  # pylint: disable=protected-access
@@ -835,7 +836,8 @@ def resolve_credentials(token: str | None = None, url: str | None = None) -> dic
 
 
 def api_backend_id(name: str) -> str:
-    """Map a local backend name to its API id.
+    """Map a local backend name to its API id -- the identifier the IonQ REST
+    API uses for a backend (the ``backend`` field in ``GET /backends``).
 
     ``ionq_qpu.forte-1``/``forte-1`` -> ``qpu.forte-1``,
     ``ionq_simulator`` -> ``simulator``, ``ionq_qpu`` -> ``qpu``.
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -798,11 +798,8 @@ def resolve_credentials(token: str | None = None, url: str | None = None) -> dic
 
 
 def _api_backend_id(name: str) -> str:
-    """Map a local backend name to its API id.
-
-    ``ionq_qpu.forte-1`` -> ``qpu.forte-1``, ``ionq_simulator`` -> ``simulator``,
-    ``forte-1`` -> ``qpu.forte-1`` (also accepts noise-model names).
-    """
+    """Local name -> API id (``ionq_qpu.forte-1``/``forte-1`` -> ``qpu.forte-1``,
+    ``ionq_simulator`` -> ``simulator``)."""
     name = name.removeprefix("ionq_")
     if name == "simulator" or name.startswith("qpu."):
         return name
@@ -812,14 +809,9 @@ def _api_backend_id(name: str) -> str:
 def get_backend_config(
     name: str, token: str | None = None, url: str | None = None
 ) -> dict:
-    """Fetch a backend's static configuration from ``GET /backends/{id}``.
-
-    Returns the raw payload -- ``qubits``, ``supported_gates``,
-    ``supported_native_gates``, ``supported_error_mitigations`` and more -- so
-    callers can derive qubit count, native gateset, and capabilities from a
-    single request instead of hardcoding them per device. Returns an empty
-    dict (with a warning) when the endpoint is unreachable, letting callers
-    fall back to sensible defaults.
+    """Fetch a backend's static config (qubits, supported/native gates, error
+    mitigations) from ``GET /backends/{id}``; ``{}`` (with a warning) if
+    unreachable so callers can fall back to defaults.
     """
     creds = resolve_credentials(token, url)
     headers = {"Authorization": f"apiKey {creds['token']}"} if creds["token"] else None
--- a/qiskit_ionq/helpers.py
+++ b/qiskit_ionq/helpers.py
@@ -140,25 +140,6 @@ GATESET_MAP = {
     "native": ionq_native_basis_gates,
 }
 
-# Native 2-qubit entangling gate by IonQ device family. Add a row when a
-# new family ships; consulted by IonQBackend._make_target().
-NATIVE_2Q_BY_FAMILY: dict[str, Literal["ms", "zz"]] = {
-    "aria": "ms",
-    "forte": "zz",
-    "tempo": "zz",
-}
-
-
-def native_2q_gate(value: str | None) -> Literal["ms", "zz"] | None:
-    """Pick "ms"/"zz" implied by a backend name or noise_model, else None."""
-    if not isinstance(value, str):
-        return None
-    value = value.lower()
-    for fam, gate in NATIVE_2Q_BY_FAMILY.items():
-        if value == fam or f"{fam}-" in value or value.endswith((f".{fam}", f"_{fam}")):
-            return gate
-    return None
-
 
 def circuit_requires_qasm3(input_circuit: QuantumCircuit) -> bool:
     """True for circuits the flat v1 gate list can't express -- mid-circuit
@@ -816,27 +797,43 @@ def resolve_credentials(token: str | None = None, url: str | None = None) -> dic
     }
 
 
-def get_n_qubits(backend, fallback=4):
-    """Get the number of qubits for a given backend."""
-    backend = backend.removeprefix("ionq_")
-    backend = (
-        backend
-        if backend == "simulator" or backend.startswith("qpu.")
-        else f"qpu.{backend}"
-    )
+def _api_backend_id(name: str) -> str:
+    """Map a local backend name to its API id.
+
+    ``ionq_qpu.forte-1`` -> ``qpu.forte-1``, ``ionq_simulator`` -> ``simulator``,
+    ``forte-1`` -> ``qpu.forte-1`` (also accepts noise-model names).
+    """
+    name = name.removeprefix("ionq_")
+    if name == "simulator" or name.startswith("qpu."):
+        return name
+    return f"qpu.{name}"
+
+
+def get_backend_config(
+    name: str, token: str | None = None, url: str | None = None
+) -> dict:
+    """Fetch a backend's static configuration from ``GET /backends/{id}``.
+
+    Returns the raw payload -- ``qubits``, ``supported_gates``,
+    ``supported_native_gates``, ``supported_error_mitigations`` and more -- so
+    callers can derive qubit count, native gateset, and capabilities from a
+    single request instead of hardcoding them per device. Returns an empty
+    dict (with a warning) when the endpoint is unreachable, letting callers
+    fall back to sensible defaults.
+    """
+    creds = resolve_credentials(token, url)
+    headers = {"Authorization": f"apiKey {creds['token']}"} if creds["token"] else None
     try:
-        return (
-            requests.get(
-                f"{resolve_credentials()['url']}/backends/{backend}", timeout=5
-            )
-            .json()
-            .get("qubits", fallback)
+        resp = requests.get(
+            f"{creds['url']}/backends/{_api_backend_id(name)}",
+            headers=headers,
+            timeout=5,
         )
+        resp.raise_for_status()
+        return resp.json()
     except Exception as exception:  # pylint: disable=broad-except
-        warnings.warn(
-            f"Failed to get qubits for {backend}: {exception}. Using {fallback}"
-        )
-        return fallback
+        warnings.warn(f"Failed to fetch config for {name}: {exception}.")
+        return {}
 
 
 def retry(
@@ -921,6 +918,6 @@ __all__ = [
     "decompress_metadata_string",
     "get_user_agent",
     "resolve_credentials",
-    "get_n_qubits",
+    "get_backend_config",
     "retry",
 ]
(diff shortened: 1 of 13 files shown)


qiskit_ionq/helpers.py
```diff
@@ -815,27 +797,35 @@ def resolve_credentials(token: str | None = None, url: str | None = None) -> dic
...
-        if backend == "simulator" or backend.startswith("qpu.")
-        else f"qpu.{backend}"
-    )
+def _api_backend_id(name: str) -> str:
+    """Local name -> API id (``ionq_qpu.forte-1``/``forte-1`` -> ``qpu.forte-1``,
+    ``ionq_simulator`` -> ``simulator``)."""
+    name = name.removeprefix("ionq_")
+    if name == "simulator" or name.startswith("qpu."):
+        return name
+    return f"qpu.{name}"
```
antalszava: minor: it feels like somehow this logic tightly belongs with the `Backend._api_backend_name` method

qiskit_ionq/helpers.py
```diff
@@ -815,27 +797,35 @@ def resolve_credentials(token: str | None = None, url: str | None = None) -> dic
...
+
+
+def get_backend_config(
+    name: str, token: str | None = None, url: str | None = None
+) -> dict:
+    """Fetch a backend's static config (qubits, supported/native gates, error
+    mitigations) from ``GET /backends/{id}``; ``{}`` (with a warning) if
```
antalszava: minor
```suggestion
    """Fetch a backend's static config (qubits, supported/native gates, supported error
    mitigation techniques) from ``GET /backends/{id}``; ``{}`` (with a warning) if
```

antalszava: minor: point about warning could be cleaner to have in new sentence or a `Warning:` section.

qiskit_ionq/ionq_backend.py
```diff
@@ -111,14 +111,27 @@ def __init__(
...
-        self._num_qubits: int = num_qubits
         self._simulator: bool = simulator
         self._max_experiments: int | None = max_experiments
         self._max_shots: int | None = max_shots
 
-        # Target (basis & connectivity)
-        self._target = self._make_target()
-        # Track noise_model for native simulator target caching
+        # Static config from GET /backends/{id}; {} offline. Skipped when
+        # num_qubits is given explicitly (test mocks) to stay network-free.
```
antalszava: It took me a few reads to understand this comment, suggest a bit of rephrasing for more clarity; `(test mocks) to stay network-free.` is still a bit unclear.

qiskit_ionq/ionq_backend.py
```diff
@@ -111,14 +111,27 @@ def __init__(
...
+        if num_qubits is None:
+            creds = self._provider.credentials
+            self._config = get_backend_config(
+                name, creds.get("token"), creds.get("url")
```
antalszava: To be sure: `creds` def has `"token"` and `"url"` at this point, right?

qiskit_ionq/ionq_backend.py
```diff
@@ -111,14 +111,27 @@ def __init__(
...
+            )
+            num_qubits = int(self._config.get("qubits", 4))
```
antalszava: Why have `num_qubits = 4` if it wasn't specified in the config?

qiskit_ionq/ionq_backend.py
```diff
@@ -111,14 +111,27 @@ def __init__(
...
+        else:
+            self._config = {}
```
antalszava: It feels a little strange that this attribute is being set based on the fact whether or not `num_qubits` was passed

qiskit_ionq/ionq_backend.py
```diff
@@ -111,14 +111,27 @@ def __init__(
...
+        self._num_qubits: int = num_qubits
+
+        # Target/coupling map resolved lazily on first access (keeps __init__
+        # network-free); _cached_noise_model drives native-sim target rebuilds.
```
antalszava: * I'd suggest moving the `_cached_noise_model` down by 3 lines to have it right above the relevant attribute;
* What does `native-sim target rebuilds` mean? imho good to rephrase EDIT: ah, a simulator with native gateset.

qiskit_ionq/ionq_backend.py
```diff
@@ -145,11 +158,14 @@ def _default_options(cls) -> Options:
 
     @property
     def target(self) -> Target | None:
+        # Native-sim target depends on the noise model; rebuild on change.
         if self._simulator and self._gateset == "native":
             current_noise_model = getattr(self.options, "noise_model", "ideal")
-            if self._cached_noise_model != current_noise_model:
+            if self._target is None or self._cached_noise_model != current_noise_model:
```
antalszava: Could we just have a block of
```py
if self._target is None:
    self._target = self._make_target()
```
right at the beginning of this method and then `self._target = self._make_target()` could be removed from the noise model block? (and no need for the `elif`)

qiskit_ionq/ionq_backend.py
```diff
@@ -161,10 +177,25 @@ def basis_gates(self) -> Sequence[str]:
         """Return the basis gates for this backend."""
         return self._basis_gates
 
+    @property
+    def supported_gates(self) -> list[str]:
+        """QIS gate names the backend accepts (API ``supported_gates``)."""
+        return list(self._config.get("supported_gates") or [])
```
antalszava: Hmm, being presented with an empty list simply on calling this property may not be clear to a user on what's going wrong - perhaps somehow either pulling the config on such a property too or having an informative (warning/maybe error) message could help?

qiskit_ionq/ionq_backend.py
```diff
@@ -283,10 +314,72 @@ def __eq__(self, other):
...
+                f"Could not fetch connectivity for {self.name}: {exc}. "
+                "Falling back to all-to-all coupling.",
+                stacklevel=2,
+            )
+            return None
+        if cal is not None and cal.connectivity:
+            return cal.connectivity
+        return None
+
+    def _resolve_coupling_map(self) -> CouplingMap:
```
antalszava: Name of the method is a verb, but returns something - how about `_get_resolved_coupling_map` or comparable?

qiskit_ionq/ionq_backend.py
```diff
@@ -283,10 +314,72 @@ def __eq__(self, other):
...
+        """Cached coupling map: a genuine connectivity subset is restricted
+        (symmetrized); a complete or missing graph stays implicit all-to-all."""
```
antalszava: This docstring is a bit hard to read for me, perhaps good to rephrase a little and elaborate on the cases mentioned.

qiskit_ionq/ionq_backend.py
```diff
@@ -283,10 +314,72 @@ def __eq__(self, other):
     def __hash__(self):
         return hash((self.name, self._gateset))
 
+    def _fetch_connectivity(self) -> list[tuple[int, int]] | None:
+        """Two-qubit pairs from the characterization; ``None`` (-> all-to-all)
+        for simulators or any failed/empty lookup."""
```
antalszava: Also: when/how does lookup fail/return as empty? good to hint on in the docstring imho.
```suggestion
        """Get the connectivity data (two-qubit pairs from the characterization).
        
        Returns ``None`` (meaning all-to-all connectivity) for simulators or any failed/empty lookup."""
```

qiskit_ionq/ionq_backend.py
```diff
@@ -283,10 +314,72 @@ def __eq__(self, other):
...
+            # n*(n-1) directed edges == complete graph -> treat as all-to-all.
+            if edges and len(edges) < n * (n - 1):
+                cmap = CouplingMap()
+                for qubit in range(n):
+                    cmap.add_physical_qubit(qubit)
+                for src, dst in sorted(edges):
+                    cmap.add_edge(src, dst)
+                self._coupling_map = cmap
+                self._restricted_coupling = True
+                return self._coupling_map
```
antalszava, line 411: Could be great to unit test this

qiskit_ionq/ionq_backend.py
```diff
@@ -283,10 +314,72 @@ def __eq__(self, other):
...
+                left, right = int(left), int(right)
+                for src, dst in ((left, right), (right, left)):
+                    if src != dst and 0 <= src < n and 0 <= dst < n:
+                        edges.add((src, dst))
+            # n*(n-1) directed edges == complete graph -> treat as all-to-all.
+            if edges and len(edges) < n * (n - 1):
+                cmap = CouplingMap()
+                for qubit in range(n):
+                    cmap.add_physical_qubit(qubit)
+                for src, dst in sorted(edges):
```
antalszava: why sort? good to comment on

qiskit_ionq/ionq_backend.py
```diff
@@ -283,10 +314,72 @@ def __eq__(self, other):
...
+    def _resolve_coupling_map(self) -> CouplingMap:
+        """Cached coupling map: a genuine connectivity subset is restricted
+        (symmetrized); a complete or missing graph stays implicit all-to-all."""
+        if self._coupling_map is not None:
+            return self._coupling_map
+
+        n = self._num_qubits
+        pairs = self._fetch_connectivity()
+        if pairs:
+            edges: set[tuple[int, int]] = set()
```
antalszava: * Could we simply pass `pairs` as the `couplinglist` arg to `CouplingMap`'s instantiation?
* If not, could building `edges` be skipped and directly building the `CouplingMap` in the loop without collecting into a set first?

qiskit_ionq/ionq_backend.py
```diff
@@ -283,10 +314,72 @@ def __eq__(self, other):
...
+                return self._coupling_map
+
+        self._coupling_map = CouplingMap.from_full(n)
+        self._restricted_coupling = False
+        return self._coupling_map
+
     def _make_target(self) -> Target:
-        """Build a Target exposing either QIS or IonQ-native gates."""
+        """Build a Target of QIS or native gates, scoping 2q gates to the
+        coupling map on restricted topologies (else all-to-all)."""
```
antalszava:
```suggestion
        coupling map on restricted topologies (else use all-to-all connectivity)."""
```

qiskit_ionq/ionq_backend.py
```diff
@@ -283,10 +314,72 @@ def __eq__(self, other):
...
         tgt = Target(num_qubits=self._num_qubits)
 
+        self._resolve_coupling_map()
+        two_q_props = (
+            {edge: None for edge in self._coupling_map.get_edges()}
+            if self._restricted_coupling and self._coupling_map is not None
+            else None
+        )
+
+        def add(instruction) -> None:
```
antalszava: why have this helper? name could be perhaps `add_instruction_to_circuit` or so; also, why some further uses of `tgt.add_instruction(MCXGate, name="mcx")` were not ported into it?

qiskit_ionq/ionq_backend.py
```diff
@@ -333,27 +426,36 @@ def _make_target(self) -> Target:
...
-                tgt.add_instruction(MSGate(phi0, phi1, theta))
+                add(MSGate(phi0, phi1, theta))
 
         # Always allow measure/reset
         for cls in (Measure, Reset):
             tgt.add_instruction(cls())
 
         return tgt
 
+    def _two_q_native_gate(self) -> Literal["ms", "zz"]:
```
antalszava, line 498: is this method being used?

qiskit_ionq/ionq_backend.py
```diff
@@ -402,7 +503,7 @@ def with_name(self, name: str, **kwargs) -> IonQSimulatorBackend:
 
 
 class IonQQPUBackend(IonQBackend):
-    """IonQ trapped-ion hardware back-ends (Aria: MS; Forte: ZZ)."""
+    """IonQ trapped-ion hardware back-ends (Aria: MS; Forte/Tempo: ZZ)."""
```
antalszava, line 557: 🥳 🥳 🥳

test/ionq_backend/test_base_backend.py
```diff
@@ -486,3 +496,101 @@ def test_forte_rzz_transpiles_to_zz(provider):
...
+)
+def test_capabilities_from_api(provider, monkeypatch, name, two_q):
+    """Qubit count, native gateset, and capability lists come from the API config."""
+    config = {
+        "qubits": 36,
+        "supported_gates": ["x", "cnot"],
+        "supported_native_gates": ["gpi", "gpi2", two_q],
+        "supported_error_mitigations": ["Debias", "Sharpen"],
+    }
+    monkeypatch.setattr(ionq_backend, "get_backend_config", lambda *a, **k: config)
```
antalszava: This assumes that `get_backend_config` is doing the API config well - that's a fundamental assumption, so good to have a test for `get_backend_config` too (unless it's being implicitly tested by an e2e test as another case, then such a unit test is valuable, but perhaps not critical)

antalszava (commented):
Awesome and very exciting 😍💪 Left some nitpicky comments and a few questions

qiskit_ionq/ionq_backend.py
splch: Covered by the tests added in the review follow-up commits: `test_coupling_map_restricted` (subset -> symmetrized edges, and the Target scopes `zz` to exactly those pairs), `test_coupling_map_complete` (complete pair list stays implicit all-to-all), `test_coupling_map_fallback` (no connectivity -> all-to-all), and `test_sim_skips_connectivity` - all in `test/ionq_backend/test_base_backend.py`.

qiskit_ionq/ionq_backend.py
splch: Yes - `_make_target` calls it in the native-gateset branch to pick `ZZGate` (Forte/Tempo) vs `MSGate` (Aria), including resolving the simulator through its configured noise model.

qiskit_ionq/helpers.py
splch: agreed they belong together, so `_api_backend_name` just delegates to `api_backend_id()` now. kept it in helpers since the provider needs the same mapping with no backend in hand (catalog keys, noise model names)

qiskit_ionq/ionq_backend.py
splch: yes, `resolve_credentials()` always returns both keys (values can be None). moot now tho, the backend doesn't touch creds anymore since the provider fetches the catalog itself

qiskit_ionq/ionq_backend.py
splch: it keeps the old `get_n_qubits(fallback=4)` behavior so offline construction still works. it's a named `_DEFAULT_NUM_QUBITS` constant with a comment now

qiskit_ionq/ionq_backend.py
splch: intentional! explicit `num_qubits` means the caller pins its own config (MockBackend in tests) and None means the catalog is the source of truth. it's a local dict read now so it's about precedence, not network. the init comment spells it out

qiskit_ionq/ionq_backend.py
splch: done! they warn now when the config is unavailable so unknown isn't mistaken for empty. went with a warning over an error so offline still works. covered by `test_capabilities_warn_offline`

qiskit_ionq/ionq_backend.py
splch: the set does three things `couplinglist` wouldn't: symmetrize (the api lists each pair once), dedupe and bounds check, and give an edge count to compare against n*(n-1). that comparison is how a complete graph (what aria/forte report) stays implicit all-to-all. it also keeps trailing isolated qubits as nodes. added a comment on it

qiskit_ionq/ionq_backend.py
splch: it scopes 2q gates to the coupling map on restricted topologies, otherwise every add repeats that check. it takes `name=` now and everything routes through it (mcx/mcphase/PauliEvolution and measure/reset). named it `add_gate` since it adds to the target, not a circuit

qiskit_ionq/helpers.py
splch: this function's gone now, the follow-ups collapsed it into one `GET /backends` catalog call (`get_backends`). the warning note got its own sentence in the new docstring

test/ionq_backend/test_base_backend.py
splch: `get_backend_config` is gone, it's one `GET /backends` fetch now. both halves are tested: `test_get_backends_success` and `test_get_backends_fallback` cover the fetch, and the provider lookup has its own tests in `test/ionq_provider/test_backend_service.py` (known id, unknown id warning, silent defaults)

splch:
thanks for the review!! replied to the open questions inline. all the doc/naming nits are applied in the follow-up commits: docstring and comment rewrites, the `_get_coupling_map` rename, single target build site, and `add_gate` takes `name=` now. the per-backend config fetches also collapsed into one `GET /backends` catalog, and the provider lookup got its own tests :slightly_smiling_face:

qiskit_ionq/ionq_provider.py
```diff
@@ -62,13 +63,29 @@ def __init__(
...
+        """Static config for ``name`` from the cached ``/backends`` catalog.
+
+        Returns ``{}`` for the generic ``ionq_qpu`` template (no real device)
+        and warns for an unknown backend when the catalog is populated.
+        """
+        api_id = api_backend_id(name)
+        config = self._catalog.get(api_id)
+        if config is None and self._catalog and api_id != "qpu":
+            warnings.warn(
+                f"Backend {api_id!r} not in the IonQ catalog; using defaults."
```
antalszava: as `backend_config` is called in the "background" it may be better to make this warning message more descriptive - what does "defaults" refer to?

qiskit_ionq/ionq_provider.py
```diff
@@ -62,13 +63,29 @@ def __init__(
         super().__init__()
         self.custom_headers = custom_headers
         self.credentials = resolve_credentials(token, url)
+        # One GET /backends call serves every backend's static config.
+        self._catalog = get_backends(self.credentials["token"], self.credentials["url"])
```
antalszava: is it def safe to index into `self.credentials` at this point?

qiskit_ionq/helpers.py
```diff
@@ -846,27 +834,39 @@ def resolve_credentials(token: str | None = None, url: str | None = None) -> dic
...
+    ``ionq_simulator`` -> ``simulator``, ``ionq_qpu`` -> ``qpu``.
+    """
+    name = name.removeprefix("ionq_")
+    if name in ("simulator", "qpu") or name.startswith("qpu."):
+        return name
+    return f"qpu.{name}"
+
+
+def get_backends(token: str | None = None, url: str | None = None) -> dict[str, dict]:
+    """Fetch the ``GET /backends`` catalog, keyed by API id.
```
antalszava, line 851: What's an API id?

qiskit_ionq/helpers.py
```diff
@@ -846,27 +834,39 @@ def resolve_credentials(token: str | None = None, url: str | None = None) -> dic
...
+
+    One request returns every backend's static config (qubit count,
+    supported/native gates, supported error-mitigation techniques), so the
+    provider can resolve them all locally instead of querying per device.
```
antalszava, line 855: Nice description 😍

qiskit_ionq/ionq_provider.py
antalszava: I think it's not the best design to make even a GET request simply when instantiating a provider - such a request should be made from the user wanting something explicitly. Personally, I would rather put it into `get_backend`: that is an explicit step where the user is asking to get the backend itself - tbh. the naming of functions is great and they are already telling this story: `get_backends` should be called in the `IonQProvider.get_backend` method instead of silently being called on the provider creation.

antalszava: edit: imho `backend_config` should become `get_backend_config` - this is literally making a GET request to get the backend config - just how its name suggests. if the `backends/` endpoint was already quired, caching can be used. `get_backend` can then depend on `backend_config` if need be.

qiskit_ionq/ionq_job.py
```diff
@@ -731,6 +731,20 @@ def _format_result_qasm3(self, shots: list):
         # Zero-padded binary so get_counts() splits by creg_sizes.
         width = header.get("memory_slots") or len(clbit_labels)
 
+        # Registers absent from the artifact fold as all-zeros; warn so a
```
antalszava: this comment took me a couple of reads, how about something like
```
# Missing registers default to all-zeros. Emit a warning because an
# omitted register (e.g. compiler-elided measurements) is not the same
# as a register that was actually measured and found to be all zeros.
```

qiskit_ionq/ionq_client.py
```diff
@@ -320,18 +320,25 @@ def get_artifact(
...
-        if "octet-stream" in res.headers.get("Content-Type", "").lower():
+        content_type = res.headers.get("Content-Type", "").lower()
+        if "octet-stream" in content_type:
             return res.content
-        return json.loads(res.text, object_pairs_hook=OrderedDict)
+        if content_type.startswith("text/"):
+            return res.text
+        try:
+            return json.loads(res.text, object_pairs_hook=OrderedDict)
+        except ValueError:
```
antalszava, line 340: when does this happeng? could this be a type error too?

qiskit_ionq/helpers.py
```diff
@@ -475,6 +457,12 @@ def _qasm3_data(circuit: QuantumCircuit) -> str:
         dumps as _qasm3_dumps,
     )
 
+    # Drop the layout so dumps emits a virtual register (qubit[n] q;), which
+    # the API requires, not physical qubits ($0).
```
antalszava: what's ` ($0)`?

qiskit_ionq/helpers.py
```diff
@@ -846,27 +834,39 @@ def resolve_credentials(token: str | None = None, url: str | None = None) -> dic
...
+        return {
+            b["backend"]: b for b in items if isinstance(b, dict) and "backend" in b
+        }
     except Exception as exception:  # pylint: disable=broad-except
-        warnings.warn(
-            f"Failed to get qubits for {backend}: {exception}. Using {fallback}"
-        )
-        return fallback
+        warnings.warn(f"Failed to fetch backends catalog: {exception}.")
+        return {}
```
antalszava: Why not strictly raise an error here? we want to make a request and it doesn't work for some reason - I think the user should be notified exactly why instead of gracefully handling it with a warning, but maybe I don't understand the full picture.

qiskit_ionq/ionq_backend.py
```diff
@@ -71,12 +71,17 @@
...
-from .helpers import GATESET_MAP, get_n_qubits, native_2q_gate, warn_bad_transpile_level
+from .helpers import GATESET_MAP, api_backend_id, warn_bad_transpile_level
 from .ionq_client import Characterization
 
 if TYPE_CHECKING:  # pragma: no cover
     from .ionq_provider import IonQProvider
 
+# Fallback qubit count when the /backends catalog has no config for a backend
+# (offline, unknown id), so a Target can still be built; real counts come from
+# the API.
```
antalszava, line 82: I'm seeing this comment after my other comment for `get_backends` - I still think an error should be raised there and then an error could be handled where we want the "graceful" handling in other places.

qiskit_ionq/ionq_backend.py
```diff
@@ -111,14 +116,27 @@ def __init__(
...
-        self._num_qubits: int = num_qubits
         self._simulator: bool = simulator
         self._max_experiments: int | None = max_experiments
         self._max_shots: int | None = max_shots
 
-        # Target (basis & connectivity)
-        self._target = self._make_target()
-        # Track noise_model for native simulator target caching
+        # Static config (qubits, gates, capabilities) comes from the provider's
+        # cached /backends catalog. An explicit num_qubits (e.g. MockBackend in
```
antalszava: imho it's more suited not to reference tests in the code for logic

qiskit_ionq/ionq_backend.py
```diff
@@ -147,11 +165,15 @@ def _default_options(cls) -> Options:
 
     @property
     def target(self) -> Target | None:
-        if self._simulator and self._gateset == "native":
-            current_noise_model = getattr(self.options, "noise_model", "ideal")
-            if self._cached_noise_model != current_noise_model:
-                self._target = self._make_target()
-                self._cached_noise_model = current_noise_model
+        # A native-gateset simulator's Target depends on the active noise model
```
antalszava:
```suggestion
        # A native-gateset simulator's target depends on the active noise model
```

qiskit_ionq/ionq_backend.py
```diff
@@ -163,10 +185,38 @@ def basis_gates(self) -> Sequence[str]:
...
 
+    def _config_list(self, key: str) -> list[str]:
+        """Read a list-valued key from the API config, warning when the config
+        is unavailable so an empty result is distinguishable from "none"."""
+        if not self._config:
+            warnings.warn(
+                f"No API config available for backend {self.name!r}; "
+                f"{key} is unknown (returning an empty list)."
+            )
+        return list(self._config.get(key) or [])
```
antalszava: we can still return an empty list and not have a warning - why only warn if the config is empty?

antalszava (changes requested):
Shaping out very nicely! 🎉 Main comment for which requesting change is when the GET API call to `backends` is being made - imho it's better to not do it silently to users, but when explicitly part of another query. Open to discussion on it nonetheless. 👍

qiskit_ionq/ionq_provider.py
splch: done! nothing fetches on init anymore. `get_backend_config` fetches the catalog once and caches it, and `get_backend` depends on it, so unknown names warn when you ask for the backend. and yes, `resolve_credentials()` always returns both keys, the values can just be None.

qiskit_ionq/helpers.py
splch: agreed, `get_backends` just raises now. `get_backend_config` catches, warns once, and caches an empty catalog so offline still works.

qiskit_ionq/ionq_backend.py
splch: yep, exactly that split: the helper raises and the provider handles the fallback.

qiskit_ionq/ionq_backend.py
splch: the warning covers the one ambiguous case, no config at all. when the entry exists, a missing key means none, not unknown, since the api omits keys that don't apply. every backend in the live catalog returns all three keys today. happy to warn on missing keys too if you prefer.

qiskit_ionq/ionq_client.py
splch: it fires when the content type is missing or unrecognized and the body isn't json, so we return the text instead of crashing. TypeError can't happen since `requests.Response.text` is always a str.

splch:
thanks for another round!! the catalog fetch is lazy now like you suggested. nothing on init, `get_backend_config` fetches once and caches, and `get_backends` raises with the provider owning the fallback. replied to the rest inline :slightly_smiling_face:

qiskit_ionq/ionq_provider.py
```diff
@@ -62,13 +63,41 @@ def __init__(
...
+            except Exception as exception:  # pylint: disable=broad-except
+                warnings.warn(f"Failed to fetch backends catalog: {exception}.")
+                self._catalog = {}
+        api_id = api_backend_id(name)
+        config = self._catalog.get(api_id)
+        if config is None and self._catalog and api_id != "qpu":
+            warnings.warn(
+                f"Backend {api_id!r} not in the IonQ catalog; it will be "
+                "built with a fallback qubit count, and its supported gates "
+                "and error mitigations will be reported as unknown."
```
antalszava:
```suggestion
                "and error mitigation options will be reported as unknown."
```

qiskit_ionq/helpers.py
```diff
@@ -852,21 +854,15 @@ def get_backends(token: str | None = None, url: str | None = None) -> dict[str,
...
-            b["backend"]: b for b in items if isinstance(b, dict) and "backend" in b
-        }
-    except Exception as exception:  # pylint: disable=broad-except
-        warnings.warn(f"Failed to fetch backends catalog: {exception}.")
-        return {}
+    resp = requests.get(f"{creds['url']}/backends", headers=headers, timeout=5)
+    resp.raise_for_status()
+    data = resp.json()
+    items = data if isinstance(data, list) else (data.get("backends") or [])
+    return {b["backend"]: b for b in items if isinstance(b, dict) and "backend" in b}
```
antalszava: minor: maybe `data` and `items` could be swapped with more descriptive names for quicker readability here

qiskit_ionq/ionq_client.py
antalszava: got it! maybe good to add a short comment on it

antalszava (approved):
Looks good! 🥇 😍

qiskit_ionq/helpers.py
splch: totally thanks! i got that added

PR #264, opened 2026-06-26
Add debiasing and symmetry verification options to `backend.run`

natestemen:
### Summary

Adds new error mitigation API, exposing debiasing and symmetry verification as booleans, and new result-aggregation methods.

### Details and comments

Two boolean flags in `backend.run`

1. `debiasing=...` and
2. `symmetry_verification=...`.

as well as a new `AggregationMethod` object available at the top level which can be passed into `job.result(aggregation=...)`.

The old methods for applying EM (`error_mitigation=qiskit_ionq.DEBIASING` and `sharpen=True`) are deprecated.

These features are documented in the README.


natestemen:
Got most of the CI to pass, but the restriction on long names for tests is tripping me up. The rules makes sense for production code, but why restrict on test names?

qiskit_ionq/ionq_client.py
```diff
@@ -259,7 +259,7 @@ def get_latest_calibration(self, backend_name: str) -> Characterization | None:
     def get_results(
         self,
         results_url: str,
-        sharpen: bool | None = None,
+        aggregation: str | None = None,
```
antalszava: Removing `sharpen` as an argument of a user-facing function is a breaking change - are we okay with that?

qiskit_ionq/error_mitigation.py
```diff
@@ -0,0 +1,116 @@
+# Copyright 2024 IonQ, Inc. (www.ionq.com)
```
antalszava: The Qiskit licensing header may also be required as in other files. Relevant for other files too.

qiskit_ionq/error_mitigation.py
```diff
@@ -0,0 +1,116 @@
...
+    """Aggregation method for results from a debiased job."""
+
+    AVERAGE = "average"
+    VOTING = "voting"
+    DNL = "dnl"
+
+
+@dataclass
+class Twirling:
+    """Two-qubit gate twirling options for debiasing."""
```
antalszava: It's a little confusing perhaps that the class is called `Twirling` (quite general), but as per the docstring it stands for two-qubit twirling. If that's true, maybe good to adjust the class name.

qiskit_ionq/error_mitigation.py
```diff
@@ -0,0 +1,116 @@
...
+        result["one_qubit_twirling"] = (
+            self.one_qubit.value
+            if isinstance(self.one_qubit, OneQubitTwirling)
+            else self.one_qubit
+        )
+        return result
+
+
+@dataclass
+class Debiasing:
```
antalszava: Perhaps good to make this `DebiasingConfig` too to match the name of the next class. it also "tells a better story" perhaps: `ErrorMitigationConfig.debiasing`'s attribute taking a `DebiasingConfig` object means that's how debiasing is being configured, otherwise it may be a bit unclear what `Debias` actually does just based on the name.

qiskit_ionq/error_mitigation.py
```diff
@@ -0,0 +1,116 @@
...
+        if self.num_variants is not None:
+            result["num_variants"] = self.num_variants
+        if self.twirling is not None:
+            result["phi_chi_twirling"] = self.twirling.to_dict()
+        return result
+
+
+@dataclass
+class ErrorMitigationConfig:
+    """Bundle of all error mitigation options, for use with
```
antalszava: very minor to match classname's terminology
```suggestion
    """Bundle of all error mitigation configurations, for use with
```

qiskit_ionq/error_mitigation.py
```diff
@@ -0,0 +1,116 @@
...
+@dataclass
+class Debiasing:
+    """Fine-grained debiasing options."""
+
+    num_variants: int | None = None
+    twirling: Twirling | None = None
+
+    def to_dict(self) -> dict:
+        """Serialize to API wire format."""
+        result = {"debiasing": True}
```
antalszava: Minor: imho it makes more sense to move this logic into the block of `if isinstance(self.debiasing, Debiasing):` within `ErrorMitigationConfig.to_dict` as that's where `result["debiasing"]` is being set in the `else` case. This can help with future maintainability.

test/helpers/test_error_mitigation_settings.py
```diff
@@ -0,0 +1,299 @@
...
+    Debiasing,
+    ErrorMitigationConfig,
+    PhiChiPattern,
+    Twirling,
+)
+from qiskit_ionq import ionq_job
+from .. import conftest
+
+
+def _simple_circuit():
```
antalszava, line 41: Imho it would be great to have integration test cases too where the new objects from the EM module are tested as a user would interact with them in Qiskit - e.g., passing them into `run` method of a backend.

qiskit_ionq/helpers.py
```diff
@@ -486,17 +487,54 @@ def _qasm3_data(circuit: QuantumCircuit) -> str:
...
+            "Use the debiasing= and symmetry_verification= kwargs on backend.run() instead.",
+            DeprecationWarning,
+        )
+        em_cfg = dict(bundle.value)
+    elif isinstance(bundle, ErrorMitigationConfig):
+        em_cfg = bundle.to_dict()
+    else:
+        em_cfg = {}
+
+    # Flat kwargs override the bundle (None means "not set", so skip)
```
antalszava: Why not simply raise an error if both
* `"error_mitigation"` and
* `"debiasing"` or `"symmetry_verification"`
are specified?

I think stricter input validation could help more than silently preferring one way of setting to another (if I understand correctly that's what's happening atm.)

qiskit_ionq/ionq_job.py
```diff
@@ -398,9 +400,21 @@ def result(
...
         """
-        # Validate args
+        # Resolve aggregation method, with sharpen as a deprecated alias.
         if sharpen is not None and not isinstance(sharpen, bool):
             warnings.warn("Invalid sharpen type")
+            sharpen = None
+
+        if sharpen is True:
+            warnings.warn(
+                "sharpen=True is deprecated. Use aggregation='voting' instead.",
```
antalszava: In the `README.md` (and wherever the deprecation is documented) it could also be mentioned when such a warning would turn into an error (i.e., by which Qiskit-IonQ release)

test/helpers/test_error_mitigation_settings.py
```diff
@@ -0,0 +1,299 @@
...
+    qc = QuantumCircuit(1, 1)
+    qc.h(0)
+    qc.measure(0, 0)
+    return qc
+
+
+def _em_settings(ionq_json):
```
antalszava: minor: `_extract_em_settings` may be a more telling name for readability

test/helpers/test_error_mitigation_settings.py
```diff
@@ -0,0 +1,299 @@
...
+def test_bundle_defaults(simulator_backend):
+    """ErrorMitigationConfig() enables both debiasing and symmetry verification."""
+    args = {"shots": 10, "error_mitigation": ErrorMitigationConfig()}
+    em_block = _em_settings(qiskit_to_ionq(_simple_circuit(), simulator_backend, passed_args=args))
+    assert em_block == {"debiasing": True, "symmetry_verification": True}
+
+
+def test_bundle_debiasing_false(simulator_backend):
+    """ErrorMitigationConfig(debiasing=False) disables debiasing."""
+    args = {"shots": 10, "error_mitigation": ErrorMitigationConfig(debiasing=False)}
```
antalszava: Since the above test cases show that passing in debiasing and sv options as flat kwargs is accepted syntax, I don't quite see the upside of the `ErrorMitigationConfig` object. It's a little easier for a user since it turns on sv by default, but they need to know this and by then they may as well simply just turn on sv manually (or if debiasing is passed as a kwarg, but sv is not, then the default could easily be to have sv on by default). Or if that's too much of a hidden behaviour, the UI of `ErrorMitigationConfig(debiasing=False)` also requires users to know the hidden default `sv=True`.

It can also confuse some users to support both the use of an `ErrorMitigationConfig` object and passing in kwargs flat - what is the difference?

antalszava (commented):
Amazing one! 🎉  main points about a potentially breaking change in a signature and a little unclear what the diff between flat kwargs and the use of `ErrorMitigationConfig` is - otherwise mostly nitpicky comments 👍

qiskit_ionq/helpers.py
antalszava: I left a relevant comment in the test cases section too, and overall these together are telling a story: if I understand correctly, a user can achieve the same thing with 2 different syntaxes. That usually can lead to confusion for them as to which one to prefer and what the difference is.

qiskit_ionq/ionq_client.py
natestemen: AFAIK the `IonQClient` object is public, but never expected for users to interact with[^1]. A user must instantiate an `IonQProvider` object which internally creates a `IonQClient` instance. I was thinking because that's the case, a breaking change here would be okay, but we could offer both kwargs side by side. I'll defer to @splch to make the call. Happy to go either way.

[^1]: In which case, should the object be made "private"?

qiskit_ionq/error_mitigation.py
natestemen: Good catch. Done in https://github.com/qiskit-community/qiskit-ionq/pull/264/commits/c47d022defa6ebef267e97d1eebb01f404cc560a.

qiskit_ionq/error_mitigation.py
natestemen: Are you suggesting that `Debiasing` does not need a `to_dict` method?

qiskit_ionq/error_mitigation.py
natestemen: Going to rename this `TwirlingConfig` since it holds data about both single and two qubit twirling options, and to be in line with the comments below regarding naming.

test/helpers/test_error_mitigation_settings.py
natestemen: Yeah good idea. Here's what I came up with for testing: https://github.com/qiskit-community/qiskit-ionq/pull/264/commits/18df94019789bcb2c9e5cbfd100cb22409b7b9c7.

qiskit_ionq/helpers.py
natestemen: Yeah happy to do that. Done in https://github.com/qiskit-community/qiskit-ionq/pull/264/commits/57711ec8be53a542b0ae8fe5db85f6a2d8fca070.

qiskit_ionq/ionq_job.py
natestemen: @splch what do you typically do here?

test/helpers/test_error_mitigation_settings.py
natestemen: The idea is that as more error mitigation techniques come online (dynamical decoupling and leakage checks coming... sometime), adding a new kwarg for every technique could clutter the `run` methods namespace. This would mean encouraging users to use the EMConfig object rather than separate techniques which also become difficult to track as we start thinking about/adding techniques that are not necessarily compatible with other techniques.

I do agree that supporting 2 paths to perform the same thing is confusing, though. I guess I'm thinking support for both is temporary and once more techniques come online, we start removing kwarg access and only use EMConfig. WDYT?

yvetteds:
Looks great!  

One followup question: @natestemen where did we land on gating the `num_variants` & `num_shots` for debiasing? Thanks!

qiskit_ionq/ionq_client.py
antalszava: Hmm, while I see that point, I'm not completely sure about it as there's the `IonQBackend.client` property that potentially returns the `IonQClient`. If someone had code that returned this object and passed the `sharpen` argument directly, that code is going to break.

The likelihood of any user having such code in practice is def a question. Even so, would be good to see how such a case falls within the stanards of deprecating user-facing functionalities.

test/helpers/test_error_mitigation_settings.py
antalszava: looks good!

test/helpers/test_error_mitigation_settings.py
antalszava: > adding a new kwarg for every technique could clutter the run methods namespace

this is a fair point :+1:

> encouraging users to use the EMConfig object rather than separate techniques which also become difficult to track as we start thinking about/adding techniques that are not necessarily compatible with other techniques.

Frankly, I think it's going to cause the most confusion to first have 2 viable ways for the same thing and then deprecate one down the line. Regardless of the decision taken, it's best to be able to do one thing from the beginning and stick to it, unless absolutely necessary to change it. My personal preference here would be to have `run` take the `error_mitigation` kwarg that either takes the `ErrorMitigationConfig` object or a dictionary (here's Claude's output on the matter which imho is great):
> error_mitigation={"some_error_mitigation_technique": True}) works if you make the config field accept `ErrorMitigationConfig | dict`, internally coercing dicts via `ErrorMitigationConfig(**d)`. Now casual users get kwargs-like ergonomics, power users get a typed, validated, autocompleted object — one underlying representation, two entry shapes, no deprecation needed.

natestemen:
After chatting with other folks we decided to remove

| item | rationale |
| ---- | --------- |
| phi chi twirling | there are specific recommendations for which approach to use on each device, so users should just get the best performing option per device. This will be enabled by default per device. |
| `num_variants` | This was being sent to the circuit service, but the service wasn't actually listening for this parameter (my fault, I thought it was). We'll stick with the default of 24 variants for >= 500 shots, and 32 for >= 1,000. |
| `ErrorMitigationConfig` | purely a design choice since there are only two bools to set now, this feels much too heavy. Can potentially include/add later. |

This leaves the API of the features we are shipping as two boolean kwargs `debiasing`/`symmetry_verification` (instead of the old `error_mitigation` enum), and then some new aggregation methods.

Ready for another round of review @splch @antalszava

qiskit_ionq/constants.py
```diff
@@ -71,6 +71,14 @@ class ErrorMitigation(enum.Enum):
     NO_DEBIASING = {"debiasing": False}
 
 
+class AggregationMethod(enum.Enum):
+    """Aggregation method for results from a debiased job."""
+
+    AVERAGE = "average"
+    VOTING = "voting"
+    DNL = "dnl"
```
antalszava, line 88: This may be a well-known abbreviation, but I don't know what it means and it may be the same for some users - maybe best to clarify in some way.

qiskit_ionq/ionq_client.py
antalszava: Reminder: this still seems to be a breaking change.

test/helpers/test_error_mitigation_settings.py
```diff
@@ -0,0 +1,253 @@
...
+import warnings
+
+import pytest
+
+from qiskit import QuantumCircuit
+
+from qiskit_ionq.helpers import qiskit_to_ionq
+from qiskit_ionq.constants import AggregationMethod
+from qiskit_ionq import ionq_job
+from .. import conftest
```
antalszava: Importing `conftest.py` is generally not a good practice. Fixtures are discovered automatically by pytest, so there's no need to import them. If these are non-fixture helpers, it's better to keep them in this test file or move them to a separate helper module rather than putting them in `conftest.py`.

*(My personal preference is in-file placement just because it helps with discoverability.)*

qiskit_ionq/helpers.py
```diff
@@ -486,17 +486,47 @@ def _qasm3_data(circuit: QuantumCircuit) -> str:
...
 
-def _build_settings(passed_args: dict, backend) -> dict[str, Any]:
-    """Build ``settings`` from ``job_settings``, merging the ErrorMitigation
-    enum into ``error_mitigation``.
+def _resolve_em_config(passed_args: dict, backend) -> dict[str, Any]:
+    """Resolve error mitigation config from passed_args and backend defaults.
+
+    Flat kwargs (``debiasing``, ``symmetry_verification``) are the primary way to
+    configure error mitigation. The legacy ``ErrorMitigation`` enum is still
+    accepted via ``error_mitigation=`` with a deprecation warning.
```
antalszava, line 494: This info should be described in public docs - perhaps even more than once. Atm. this is not clear from public docstrings and/or the readme file.

antalszava (commented):
🎉 just a few minor points to clear from my side still, looking sharp 💯

qiskit_ionq/constants.py
natestemen: Good catch. I added some general docs and a link to the paper it's defined in in https://github.com/qiskit-community/qiskit-ionq/pull/264/commits/05d8a7df71f526ad7bc131ef471a5d4ce5b16a13, but IMO a paper isn't necessarily good docs. We should document this on the ionq official docs as well.

test/helpers/test_error_mitigation_settings.py
natestemen: Moved to utils in https://github.com/qiskit-community/qiskit-ionq/pull/264/commits/e3e819d637b78c996bdd9663f8b7333b5baa18b3. Let me know if this is a good solution.

antalszava: lgtm :+1:

qiskit_ionq/constants.py
antalszava: sounds good to me 👍 the docstring now is definitely a really good base

antalszava (approved):
Looks good to me 🥇 CI checks seem to be failing, otherwise this is great 💯

natestemen:
CI failure is related to long tests names (original comment: https://github.com/qiskit-community/qiskit-ionq/pull/264#issuecomment-4806118158). @splch to comment before making decision.

splch:
nice! i have a few notes:

yeah the test name length lint is tedious. just bump function-rgx to {2,50} in this PR and keep the names

one ask can you update the PR description before merge - we squash, so it becomes the commit message, and it still mentions ErrorMitigationConfig, Twirling, etc. from the earlier design.

The API shape landed in a great place. i left some small inline comments below

qiskit_ionq/constants.py
```diff
@@ -85,4 +102,10 @@ class ResultFormat(str, enum.Enum):
     PROBABILITIES_V2 = "ionq.result.probabilities.json.v2"
 
 
-__all__ = ["APIJobStatus", "JobStatusMap", "ErrorMitigation", "ResultFormat"]
+__all__ = [
```
splch, line 105: can you also expose AggregationMethod from the top-level __init__.py?

qiskit_ionq/helpers.py
```diff
@@ -486,17 +486,47 @@ def _qasm3_data(circuit: QuantumCircuit) -> str:
...
         "error_mitigation"
     )
-    if isinstance(error_mitigation, ErrorMitigation):
+
+    em_cfg: dict[str, Any] = {}
+    if isinstance(legacy_enum, ErrorMitigation):
+        warnings.warn(
+            f"Passing ErrorMitigation.{legacy_enum.name} is deprecated. "
+            "Use the debiasing= and symmetry_verification= kwargs on backend.run() instead.",
+            DeprecationWarning,
```
splch, line 507: add `stacklevel=2` here

test/helpers/test_error_mitigation_settings.py
```diff
@@ -0,0 +1,269 @@
...
+
+# ---------------------------------------------------------------------------
+# result() aggregation kwarg
+# ---------------------------------------------------------------------------
+
+
+def test_default_no_aggregation_param(mock_backend, requests_mock):
+    """result() with no aggregation kwarg sends no aggregation query param."""
+    job = _setup_job(mock_backend, requests_mock)
+    assert job.result() is not None
```
splch, line 181: `assert "aggregation" not in requests_mock.last_request.qs`

qiskit_ionq/ionq_client.py
```diff
@@ -279,11 +282,18 @@ def get_results(
         Returns:
             dict: A :mod:`requests <requests>` response :meth:`json <requests.Response.json>` dict.
         """
+        if sharpen is not None:
+            warn(
+                "The sharpen parameter is deprecated. Use aggregation='voting' instead.",
```
splch: the message only makes sense for `sharpen=True`. False isn't voting so maybe just: `"sharpen is deprecated; use aggregation= instead."`

splch (approved):

qiskit_ionq/constants.py
natestemen: done in https://github.com/qiskit-community/qiskit-ionq/pull/264/commits/157d4a6833f67e45869c1b75221dab4f49db57c8.

qiskit_ionq/helpers.py
natestemen: added in https://github.com/qiskit-community/qiskit-ionq/pull/264/commits/a03986b9faa8a7e8f1592df4cdb738e11a35af23. I haven't used this kwarg much so I wasn't totally sure if we are always aware of where these functions get called from, but handy feature!

qiskit_ionq/ionq_client.py
natestemen: good catch. done in https://github.com/qiskit-community/qiskit-ionq/pull/264/commits/f9a41aeee2d9e0a272f7c4c2fb24d90d385d5195.

natestemen:
@splch any idea why the tests never ran?

splch:
> @splch any idea why the tests never ran?

no mine also werent running :/

natestemen:
Oh it's likely the infinite hang caused by qiskit 2.5 🤦🏻

splch:
> Oh it's likely the infinite hang caused by qiskit 2.5 🤦🏻 

oh 🤣 yeah... that

natestemen:
@splch CI passing now that we have qiskit pinned (and i fixed a wording bug)!

Issue #210, opened 2024-12-09
Transpile using the native gate set of Forte 1


sebastianvromero:
### Information

- **Qiskit IonQ Provider version**: 0.5.8
- **Qiskit version**: 1.1.0

- **Python version**: 3.9.6
- **Operating system**: macOS Sequoia v15.1.1

### What is the current behavior?

Using `transpile()` from qiskit, I am trying to transpile a given circuit into the native set of gates of IonQ Forte 1, which considers [ZZ gates instead of MS gates](https://ionq.com/docs/getting-started-with-native-gates), as IonQ Aria platforms have.

### Steps to reproduce the problem

```python3
from qiskit_ionq import IonQProvider
from qiskit import QuantumCircuit, transpile

provider = IonQProvider(token='your token')
native_simulator = provider.get_backend('ionq_simulator', gateset='native')
native_simulator.set_options(noise_model='forte-1')

qc = QuantumCircuit(2)
qc.rzz(1, 0, 1)

transpiled_qc = transpile(qc, backend=native_simulator, optimization_level=3)
print(transpiled_qc.count_ops()) # OrderedDict([('gpi2', 10), ('ms', 2), ('gpi', 1)])
```

### What is the expected behavior?

Instead of MS gates, the transpiler should be aware of the backend chosen and transpile accordingly. If Forte backend is chosen, use ZZ as entangling gates instead.

### Suggested solutions

The `provider.get_backend()` function could include a new input which is the device selected. 
- If `device='aria-1'`: use GPi, GPi2 and MS gates.
- If `device='forte-1'`: use GPi, GPi2 and ZZ gates.

PR #266, opened 2026-07-08
fix: transpile deadlock on qiskit>=2.5 for native-gateset backends

splch:
## Problem

`transpile(qc, backend=<native backend>)` hangs forever on qiskit 2.5.0 at `optimization_level>=2` — including the default. CI first hit this in #264; `main` is equally affected.

Qiskit 2.5.0 made `Optimize1qGatesDecomposition` multithreaded (Qiskit/qiskit#15567). With Python-defined gates (`GPIGate`/`GPI2Gate`) in the target, the main thread enters rayon **holding the GIL** while worker threads block in `PyGILState_Ensure` trying to fetch gate matrices → deadlock (0% CPU, uninterruptible).

## Fix

Python gates must not reach the optimizer, so native targets now advertise **exact standard-gate proxies** — `r`/`rz` plus `rzz` (`zz(θ) ≡ rzz(2πθ)`) or `rxx` (`ms(0,0,θ) ≡ rxx(2πθ)`). These live in Qiskit's Rust core, so the whole preset pipeline runs GIL-free and can genuinely optimize.

A new `ionq_native` scheduling-stage plugin (auto-selected via `IonQBackend.get_scheduling_stage_plugin()`) converts the optimized circuit back to `gpi`/`gpi2`/`ms`/`zz`, exactly (global phase included, numerically verified per 1q run).

**User contract is unchanged**: `transpile()` still returns native gates — usually far fewer (the #210 regression test now yields 1 `zz` instead of 29 gates).

Also:
- `TrappedIonOptimizerPlugin` skips the preset pipeline when it has no compilation target (same deadlock).
- New `ms → rz·rxx·rz` equivalence; native circuits now also translate *out* to QIS backends.
- `u_gate_equivalence` records its previously-missing global phase `(φ+λ)/2 − π/2`.

## Testing

Full suite passes on **qiskit 2.4.2 and 2.5.0** (the previously-hanging `test_forte_rzz_transpiles_to_zz` runs in 0.02 s). Optimizer-plugin depth goldens are regenerated as upper bounds since exact depths track the qiskit version. New tests cover the symbolic-parameter path, the 1q native synthesis, and equivalence-rule exactness.

The underlying bug (entering rayon while holding the GIL) still needs an upstream Qiskit issue — any custom-gate target hits it.

### DIFF
--- a/qiskit_ionq/ionq_backend.py
+++ b/qiskit_ionq/ionq_backend.py
@@ -49,6 +49,7 @@ from qiskit.circuit.library import (
     MCPhaseGate,
     MCXGate,
     PhaseGate,
+    RGate,
     RXGate,
     RXXGate,
     RYGate,
@@ -69,7 +70,6 @@ from qiskit.circuit.library import (
 from qiskit.providers import BackendV2 as Backend, Options
 from qiskit.transpiler import Target, CouplingMap
 
-from qiskit_ionq.ionq_gates import GPIGate, GPI2Gate, MSGate, ZZGate
 from . import ionq_equivalence_library, ionq_job, ionq_client, exceptions
 from .helpers import GATESET_MAP, get_n_qubits, native_2q_gate, warn_bad_transpile_level
 from .ionq_client import Characterization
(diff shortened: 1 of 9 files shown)


antalszava:
@splch with an upstream fix in Qiskit 2.5.1 and https://github.com/qiskit-community/qiskit-ionq/pull/272 being merged, okay to close this PR?

splch:
> @splch with an upstream fix in Qiskit 2.5.1 and #272 being merged, okay to close this PR?

yes! thanks for the reminder :)

PR #267, opened 2026-07-09
chore: migrate to modern Python tooling (pyproject.toml + uv)

### LINKED ISSUE #265: Migrate to modern Python project tooling
## Background

This project predates many now-standard Python packaging conventions and uses a scattered configuration setup that creates unnecessary maintenance overhead.

### Current state

- Build: setup.py + setup.cfg + a minimal pyproject.toml
- Dependencies: Three separate files (requirements.txt, requirements-test.txt, requirements-docs.txt)
- Version: Custom version.py with manual git subprocess calls
- Test runner: tox + tox.ini
- Linting: pylint + isort, each configured separately
- CI: pip install ... tox tox-gh-actions in every job

#### Proposed changes

Consolidate into pyproject.toml
- Move all setup.py / setup.cfg metadata into [project] in pyproject.toml
- Replace the three requirements*.txt files with [dependency-groups] (test, docs)
- Move tool config from tox.ini into pyproject.toml
- Replace version.py (which shells out to git) with setuptools-scm or a plain static string updated at release — the current approach is fragile and adds ~80 lines of boilerplate
- Replace pylint + isort with ruff
- Adopt uv in CI

#### Update CONTRIBUTING.md
- Remove the reference to python setup.py test (deprecated setuptools interface)
- Document how to install the project and its dev dependencies with uv
- Document how to run tests (uv run pytest) and the linter (uv run ruff check)

@splch wdyt? At minimum we need to update the contributing file since there is no info on how to set up a development environment here.

natestemen:
Closes #265.

This migrates the project to modern Python packaging and tooling, consolidating six config files (`setup.py`, `setup.cfg`, `tox.ini`, `.pylintrc`, `MANIFEST.in`, three `requirements*.txt`) into `pyproject.toml` and adopting uv across CI and the contributor workflow.

## What changed

**Packaging** (2d2a1ae)
- All metadata moved from `setup.py`/`setup.cfg` into `[project]`, with an SPDX license expression
- The three `requirements*.txt` files became PEP 735 `[dependency-groups]` (`test`, `docs`, and a `dev` group that pulls in both)
- pytest config from `setup.cfg` + `tox.ini` merged into `[tool.pytest.ini_options]`
- Dropped the `importlib-metadata` backport in favor of stdlib `importlib.metadata`
- `uv.lock` committed for reproducible CI installs

**Versioning** (6a6df1f, 6b79b6b)
- `version.py` (~100 lines shelling out to git at import time) replaced with setuptools-scm; the version derives from the existing `v*` tags, so tagged releases need no version-bump commit, and dev installs get the same `1.1.2.dev5+g<sha>` flavor of version as before
- `__version__` and the Sphinx `release` string now come from installed-distribution metadata

**Lint config** (b082b44, 682469b)
- `.pylintrc` translated into `[tool.pylint]`, carrying only non-default values; verified for parity (zero diagnostics before and after)
- `tox.ini` deleted; the CI matrix already covers multi-Python, so tox was pure indirection
- ruff's excludes moved from the pre-commit hook into `[tool.ruff]`
- the license-header checker now skips the setuptools-scm-generated `_version.py`

**CI** (2ba7fa6)
- All workflows use `astral-sh/setup-uv` with locked installs of only the dependency group each job needs; release builds with `uv build` (with `fetch-depth: 0` so setuptools-scm sees the tag)

**Docs** (0e4b173)
- CONTRIBUTING.md gains a real dev-environment section (`uv sync`, `uv run pytest`, lint/format commands, pre-commit setup); the deprecated `python setup.py test` instructions are gone from CONTRIBUTING.md and README.md

## Heads-up: qiskit 2.5.0 regression (pre-existing, not caused by this PR)

While verifying, the test suite hung at `test_forte_rzz_transpiles_to_zz`: **qiskit 2.5.0 (released 2026-07-02) deadlocks when transpiling to the IonQ native target at `optimization_level=3`** (the main thread parks forever in a rayon lock inside `_accelerate`). This reproduces identically on unmodified `main` with the old pip-based setup — the last green CI run on `main` (June 5) simply predates the 2.5.0 release, so `main`'s next CI run would hang too. qiskit 2.4.2 is fine.

This PR constrains the locked dev/CI environment to `qiskit<2.5` via `[tool.uv] constraint-dependencies` (published metadata still allows `qiskit>=2.0.0`). This deserves its own issue — both to remove the constraint once fixed and probably to report upstream to Qiskit.

## Deviation from the issue

The issue proposed replacing pylint with ruff. **Pylint is intentionally kept** (alongside ruff, which already runs via pre-commit): its `docparams`/`docstyle` extensions validate docstring parameter lists against signatures, which ruff's stable rules don't cover yet. Its config now lives in `pyproject.toml` like everything else.

## Verification

- `uv run pytest`: 417 passed, 89.81% branch coverage (above the 80% floor) on py3.13, plus a py3.10 smoke run against the built wheel
- `uv run pylint -rn qiskit_ionq test`: 10.00/10, zero diagnostics (parity with `.pylintrc`)
- `uv run pre-commit run --all-files`: all hooks pass (ruff, ruff-format, mypy, zizmor)
- `uv build`: wheel contains `py.typed` + generated `_version.py`, excludes `test/`; installed into a fresh venv and smoke-tested (`IonQProvider`, gates, `__version__`)
- `uv run make html`: docs build succeeds and the page title shows the scm-derived version

🤖 Generated with [Claude Code](https://claude.com/claude-code)


CONTRIBUTING.md
```diff
@@ -27,16 +50,29 @@ When you're ready to make a pull request, please make sure the following is true
...
+
+### Linting and formatting
+
+Continuous integration runs [pylint](https://pylint.readthedocs.io/) as well
+as [ruff](https://docs.astral.sh/ruff/) (via pre-commit) over the codebase.
+To run the same checks locally:
 
 ```bash
-pytest [pytest-args]
+uv run pylint -rn qiskit_ionq test
```
antalszava, line 66: This command is giving some formatting issues atm. for the repo and it's not the one that's being  used in the CI (see line 46 of `ci.yml`):

```
          # Tests lint with the same config, minus the 50-char cap on function names:
          uv run --no-sync pylint -rn --function-rgx='[a-z_][a-z0-9_]+$' test

```

pyproject.toml
```diff
@@ -1,6 +1,159 @@
...
+exclude = ["test*"]
+
+[tool.setuptools.package-data]
+qiskit_ionq = ["py.typed"]
+
+[tool.setuptools_scm]
+version_file = "qiskit_ionq/_version.py"
+
+[tool.ruff]
+extend-exclude = ["docs/conf.py", "tools/verify_headers.py"]
```
antalszava, line 75: Note from Claude:

> ruff format silently lost coverage of two files. On main, the exclude for docs/conf.py and tools/verify_headers.py applied only to the ruff lint hook; moving it to top-level [tool.ruff] extend-exclude makes ruff format skip them as well (the hook passes --force-exclude). Nothing breaks today (both are format-clean under the pinned ruff 0.8.2), but future edits to those files go unformatted. If lint-only exclusion was the intent, use [tool.ruff.lint] exclude instead.

pyproject.toml
```diff
@@ -1,6 +1,159 @@
...
+    "ruff",
+    { include-group = "test" },
+    { include-group = "docs" },
+]
+
+[tool.uv]
+# qiskit 2.5.0 deadlocks when transpiling to the IonQ native target at
+# optimization_level=3 (e.g. test_forte_rzz_transpiles_to_zz); keep the
+# locked dev/CI environment on 2.4.x until that is fixed upstream.
+constraint-dependencies = ["qiskit<2.5"]
```
antalszava: this may not be relevant anymore with https://github.com/qiskit-community/qiskit-ionq/pull/268 merged and `[project] dependencies` having the same pin, right?

pyproject.toml
```diff
@@ -1,6 +1,159 @@
...
+    "Intended Audience :: Developers",
+    "Intended Audience :: Science/Research",
+    "Operating System :: Microsoft :: Windows",
+    "Operating System :: MacOS",
+    "Operating System :: POSIX :: Linux",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
```
antalszava, line 26: minor and may be out of scope: what about 3.14? qiskit supports that (at least on their GH `main`)

pyproject.toml
antalszava: Also: was this resolved in 2.5.1 (like in https://github.com/Qiskit/qiskit/pull/16592)?

pyproject.toml
```diff
@@ -1,6 +1,159 @@
...
+constraint-dependencies = ["qiskit<2.5"]
+
+[tool.setuptools.packages.find]
+exclude = ["test*"]
+
+[tool.setuptools.package-data]
+qiskit_ionq = ["py.typed"]
+
+[tool.setuptools_scm]
+version_file = "qiskit_ionq/_version.py"
```
antalszava, line 72: What I'm gathering is that this file is not being used anywhere in this repo atm. - is that true?

As far as my findings go, `from importlib.metadata import version … version("qiskit-ionq")`  reads installed-package metadata. If that's true then generating `"qiskit_ionq/_version.py"` is not necessary and the `[tool.setuptools_scm]` section should be empty for simplicity.

CONTRIBUTING.md
```diff
@@ -27,16 +50,29 @@ When you're ready to make a pull request, please make sure the following is true
...
+uv run ruff check
+uv run ruff format
```
antalszava, line 68: Running this on the repo atm. is producing changes.

natestemen:
Before addressing the changes here, I think it's worth strategizing. From what I recall, @splch wanted to move the repo to use ionq-core internally before this. I don't remember the reason for that, however. Spencer can you provide your thoughts?

natestemen:
Just rebased so this is up to date, but would still like @splch's comments before we push on this.

